PR #13, opened 2020-12-10
chore: add navbar selectLanguageName

Yue-plus:
无标题


meteorlxy:
Ah nice catch!

I changed the default options assignment of theme-default, which broke this.

Thanks!

PR #28, opened 2021-01-01
docs: declare chainMarkdown is removed

Mister-Hope:


meteorlxy:
It's even not documented in v1 docs 😅

meteorlxy:
Oh damn, it's documented but I missed orz

PR #59, opened 2021-02-12
docs: add next flag

bencodezen:
- Add config to make it easier to deploy
- Signify that it's the next repo so users aren't confused


meteorlxy:
Is the next flag necessary? I saw [vue-router-next](https://next.router.vuejs.org/) and [vuex-next](https://next.vuex.vuejs.org/) don't have the flag

PR #80, opened 2021-03-25
fix(theme-default): `scrollBehavior` will be executed after the transition ends

### LINKED ISSUE #77: [Bug report] navigate to anchor on another page does not work
### Description

Navigation to anchor on another page does not work. In dev mode, the browser will receive a warning message from vue-router. 

```
  Couldn't find element using selector "#plugins" returned by scrollBehavior.
```

### Steps to reproduce

- Reproduction link / repo: https://vuepress2.netlify.app/guide/plugin.html

### Expected behavior

Navigation to anchor 

### Screenshots

vuepress-next

### Environment info

- Browser: Chrome 89.0.4389.90

John60676:
Fixes #77

### DIFF
--- a/packages/@vuepress/theme-default/src/clientAppEnhance.ts
+++ b/packages/@vuepress/theme-default/src/clientAppEnhance.ts
@@ -3,12 +3,31 @@ import Badge from './components/global/Badge.vue'
 import CodeGroup from './components/global/CodeGroup.vue'
 import CodeGroupItem from './components/global/CodeGroupItem.vue'
 import OutboundLink from './components/global/OutboundLink.vue'
+import { scrollWaiter } from './utils'
 
 import './styles/index.scss'
 
 declare const DOCSEARCH_PROPS: unknown
 
-export default defineClientAppEnhance(({ app }) => {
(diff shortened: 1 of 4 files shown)


meteorlxy:
Sorry I didn't notice your PR :sweat:.

This has been fixed in 4b8d0cff2d7fa3d74d69d551976a3b12263e6124.

Thank you all the same :heart: :+1:

John60676:
@meteorlxy 

It doesn't matter, your solution is better 👍. There is another problem here. When the website is opened for the first time, the anchor will still fail to jump. Adding `useScrollPromise().pending()` in `defineClientAppEnhance` can solve this problem.

meteorlxy:
@John60676 

I think the initial `scrollPromise.pending()` does not help, because the transition will not be triggered on initial load, and the Promise will be pending forever. 🤔 

It should not happen in build mode as all the anchors have already been pre-rendered.

John60676:
> It should not happen in build mode as all the anchors have already been pre-rendered.

Yes, you are right, this will not happen in build mode.

PR #88, opened 2021-03-30
docs: update guide/plugin.md

Yue-plus:


meteorlxy:
Thanks, but I don't think these changes are necessary, and we have to revert / remove them after v2 release.

PR #96, opened 2021-04-04
docs: add npm links for plugins

Yue-plus:


meteorlxy:
I think the `palette` plugin is missed

PR #120, opened 2021-04-21
docs: add example for registering non-SSR-friendly global plugins

seanwuapps:
I got build error of `window is not defined` whenever I register a plugin, couldn't really find any docs on it. 

Found an example in the google analytics plugin that uses `__SSR__` and `__DEV__` got it working together with `await import`

### DIFF
--- a/docs/reference/client-api.md
+++ b/docs/reference/client-api.md
@@ -84,6 +84,27 @@ export default defineClientAppEnhance(({ app, router, siteData }) => {
 })
 ```
 
+::: tip
+
+Environment variables `__SSR__` and `__DEV__` are availalbe in the callback function.
+
+:::
+
+You can use async import to load any global features that are not SSR-friendly:
+
+```ts
+export default defineClientAppEnhance(({ app, router, siteData }) => {
+  // ...
+  if (!__SSR__) {
+    // register client-side plugins with async imports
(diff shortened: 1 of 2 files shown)


meteorlxy:
They are not only available in the client app enhance files. You can use them in every client files.

I'm planning to add this to the docs of client-api, too.

seanwuapps:
@meteorlxy happy to move this to a different place, also happy to reject this if there's plan to cover it elsewhere.
Anyways, let me know, happy to help

meteorlxy:
Closing this because of 913ca05630c67f782366e9c03373fc86afed86e5 and 27572017adb4017f1b27552e59e4bed9afd1f454.

Thanks for you help :heart:

iamanewofkotlin:
If someone encountering as the same issue is.

NOTE: 

The recently ``vuepress-next`` had replace ``global variables``. 
Recently ``global variables`` list below, and your can see where ``node_modules/@vuepress/client/types.d.ts`` to track correctly code as well.

```
declare const __VUEPRESS_VERSION__: string
declare const __VUEPRESS_DEV__: boolean
declare const __VUEPRESS_SSR__: boolean
declare const __VUE_HMR_RUNTIME__: Record<string, any>

```

PR #121, opened 2021-04-24
fix: error badge

vanyauhalin:


meteorlxy:
This is an intended breaking change, to align with the custom containers `::: danger`

See https://vuepress2.netlify.app/guide/migration.html#default-theme-change.

PR #122, opened 2021-04-24
feat(markdown): support externalIcon in config and frontmatter

Shigma:
- [x] implementation
- [x] add unit tests
- [x] documentation


meteorlxy:
What about `externalIcon?: string | false`?

This would allow users to specify a different icon instead of `<OutboundLink />`.

Shigma:
Well I'm not sure whether `string` is necessary.

meteorlxy:
Missing docs for frontmatter, i.e. `reference/frontmatter.md`

PR #128, opened 2021-05-03
fix(theme-default): fix resolving children of non-group sidebar item

recoluan:
fix failure to resolve string type file path in SidebarItem.


meteorlxy:
Would you please provide an example what this PR fixes?

meteorlxy:
I have noticed this one. 🤦

However, I'm going to refactor this file, so closing this PR.

Thanks anyway

PR #144, opened 2021-05-20
fix(markdown): decode assets link to ensure bundler can find the file

Mister-Hope:
This is an issue in all VuePress versions.

```md
![中国](./中国.png)
```

should be allowed. But `markdown-it` 's default behavior will encode it to `./%E4%B8%AD%E5%9B%BD.png`, so that  bundler won't be able to find the image file.

I made a fix in V1 by my `vuepress-plugin-md-enhance`, but I think V2 can support it.

### DIFF
--- a/packages/@vuepress/markdown/src/plugins/assetsPlugin.ts
+++ b/packages/@vuepress/markdown/src/plugins/assetsPlugin.ts
@@ -22,25 +22,27 @@ export const assetsPlugin: PluginWithOptions<AssetsPluginOptions> = (
   md.renderer.rules.image = (tokens, idx, options, env: MarkdownEnv, self) => {
     const token = tokens[idx]
 
-    // get the image link and decode link to the origin one
-    // so that bundler can find the file correctly
+    // get the image link
     const link = token.attrGet('src')
 
-    if (link) {
-      if (/^\.{1,2}\//.test(link) && env.filePathRelative) {
-        // if the link is relative path, and the `env.filePathRelative` exists
-        // add `@source` alias to the link
-        const resolvedLink = `${relativePathPrefix}/${path.join(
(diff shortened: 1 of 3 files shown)


packages/@vuepress/markdown/src/plugins/assetsPlugin.ts
```diff
@@ -1,4 +1,5 @@
 import type { PluginWithOptions } from 'markdown-it'
+import { decode } from 'mdurl'
```
meteorlxy, line 2: `mdurl` is not listed in the `dependencies` of markdown package.

We should add it explicitly.

meteorlxy (changes requested):

packages/@vuepress/markdown/src/plugins/assetsPlugin.ts
Mister-Hope: Got it, I forgot adding because it's already on the markdown-it toolchain.

meteorlxy:
I just added more test cases, and I think we should make all those cases passed

Mister-Hope:
I don't think alias is needed, as I am not thinking of any situation which a person should define a alia containing chinese. While a markdown folder likely to contain chinese images,(like harmony os official docs)

Even if we resolve those alias begin with @, alias are not forced to begin with `@`.

The only way to fit this is to decode back every link not begin with `/` , comparing with decoding every relative links now.

meteorlxy:
Maybe just decode the original `src` is enough. No need to add to many codes

meteorlxy (changes requested):

meteorlxy (approved):
Made some refactor. Thanks for your contribution :heart:

PR #150, opened 2021-05-20
fix(theme-default): handle rotate events on iPad

Mister-Hope:
On devices like iPad, the viewport will change during rotate, but the `resize` event won't be triggered.

Addtional `orientationchange` event should be listened to ensure navbar render correctly


meteorlxy:
It's marked as "Deprecated" in MDN, but it seems that there is not any alternative for it? 🤔

Mister-Hope:
It's the only way to directly fit screen rotating as far as I know.

MDN do mark it as `Deprecated`, but I can not found any deprecated info on whatwg. A lot of people are using it now.

Other h5 games may use `devicemotion`, but it's not even stardard.

PR #154, opened 2021-05-21
style(theme-default): remove extra template

Mister-Hope:


meteorlxy:
Well, this is kind of coding style thing. Similar to 

```js
if (foo) {
  str = 'foo'
} else {
  str = 'bar'
}
```

vs.

```js
if (foo) str = 'foo'
else str = 'bar'
```

Issue #161, opened 2021-05-23
[Feature request] support pnpm


xinpingwang:
### Description

[pnpm](https://pnpm.io/), yet another node package manager. which creates a non-flat node_modules, so code has no access to arbitrary packages. 

In `vuepress` site project, `vuepress` is the only exposed dependent. Thus all dependencies will installed under `vuepress/node_modules` directory. this cause two problems:

1. loaders could not found by webpack, as webpack default [context](https://webpack.js.org/configuration/entry-context/#context) is `process.cwd()`, aka project root.
2. some temp files at `docs/.vuepress/.temp/internal`, created by `@vuepress/core`, need to import `vue` and `@vuepress/client`, will cause build error.

### Proposed Solution

1. specify wepack context to `@vuepress/bundler-webpack`, as it implement the compile logic and all webpack related dependencies are added by this package
2. temp files using absolute path to import dependencies.

meteorlxy:
- Webpack context and temp files may not be the only things need to be changed. We'll have extra maintenance burden to always be aware of compatibility with a specific package manager.
- It's more like an issue of pnpm if our package could work well with other package managers.
- We already put a tip for pnpm users:

https://github.com/vuepress/vuepress-next/blob/d3e5409246a47edae93209c9ce5dd2614e14c936/docs/guide/getting-started.md#L9

xinpingwang:
Even don't plan to support `pnpm`, I still think  import `vue` and `@vuepress/client` from temp file is not a good behavior, as those package are not declared as dependencies in `vuepress` site project's package.json. If you'd like, I will create a pull request to fix that **arbitrary** `import`.

wxsms:
the fact is that, as for now, https://github.com/vuejs/vue-next is using pnpm.

kissu:
> * Webpack context and temp files may not be the only things need to be changed. We'll have extra maintenance burden to always be aware of compatibility with a specific package manager.
> * It's more like an issue of pnpm if our package could work well with other package managers.
> * We already put a tip for pnpm users:
> 
> https://github.com/vuepress/vuepress-next/blob/d3e5409246a47edae93209c9ce5dd2614e14c936/docs/guide/getting-started.md#L9

This one works perfectly fine on my side. Can benefit of PNPM with Vuepress as of today! 🎊

rxliuli:
[...]

PR #162, opened 2021-05-23
feat: webpack context config and using absolute import in generated temp files (#161)

xinpingwang:
#161


meteorlxy:
See https://github.com/vuepress/vuepress-next/issues/161#issuecomment-846936397.

PR #163, opened 2021-05-24
feat(theme-default): improve a11y of CodeGroup

Mister-Hope:


Mister-Hope:
This makes sure that CodeGroup can be visited through tab, and a outline style can be used on the tabs to show the active status.

Aria labels are added, and users can use ← and → to navigate between tabs if the tabs are focused.

packages/@vuepress/theme-default/src/client/components/global/CodeGroup.ts
```diff
@@ -63,12 +64,36 @@ export default defineComponent({
                 h(
                   'button',
                   {
+                    ref: (element) => {
+                      if (element)
+                        tabRefs.value[i] = element as HTMLButtonElement
```
meteorlxy, line 69: Wrap with `{ }`

packages/@vuepress/theme-default/src/client/styles/code-group.scss
```diff
@@ -34,6 +34,10 @@
   outline: none;
 }
 
+.code-group__nav-tab:focus-visible {
+  outline: 1px solid rgba(255, 255, 255, 0.9);
```
meteorlxy, line 38: Is this possible to re-use some existing color variables?

meteorlxy (changes requested):

packages/@vuepress/theme-default/src/client/styles/code-group.scss
Mister-Hope: Since it is in a dark background, I just copy the color from L29.

And other outline is ok to follow default value, so I don't think any need to add a variable

packages/@vuepress/theme-default/src/client/components/global/CodeGroup.ts
```diff
@@ -63,12 +64,37 @@ export default defineComponent({
...
+                      }
+                    },
                     class: `code-group__nav-tab${
                       i === activeIndex.value
                         ? ' code-group__nav-tab-active'
                         : ''
                     }`,
+                    ariaPressed: i === activeIndex.value,
+                    ariaExpanded: i === activeIndex.value,
```
meteorlxy, line 78: `i === activeIndex.value` expression has been used 3 times here. We'd better extract it to a variable

packages/@vuepress/theme-default/src/client/components/global/CodeGroup.ts
```diff
@@ -63,12 +64,37 @@ export default defineComponent({
...
+                        if (i + 1 < items.length) {
+                          activeIndex.value = i + 1
+                          tabRefs.value[i + 1].focus()
+                        }
+                      } else if (event.key === 'ArrowLeft') {
+                        event.preventDefault()
+                        if (i - 1 >= 0) {
+                          activeIndex.value = i - 1
+                          tabRefs.value[i - 1].focus()
+                        }
```
meteorlxy, line 95: We could allow navigating from the last item to the first item directly, and vice versa.

meteorlxy (changes requested):

packages/@vuepress/theme-default/src/client/components/global/CodeGroup.ts
meteorlxy: In addtion, we'd better extract those navigation methods to the setup function.

For example:

```ts
const activeIndex = ref(-1)
const tabRefs = ref<HTMLButtonElement[]>([])
const activateNext = () => {
  // ...
}
const activatePrev = () => {
  // ...
}
```

meteorlxy (approved):

PR #164, opened 2021-05-24
fix(theme-default): show header anchors when being focused

Mister-Hope:


Mister-Hope:
This fix makes anchor visable when being focused by tab.

PR #168, opened 2021-05-26
fix(theme-default): fix long inlinecode breaks the layout

Mister-Hope:


Mister-Hope:
This fill fix issues like this https://waline.js.org/client/basic.html#emojicdn

image

meteorlxy:
@Mister-Hope Need to resolve the conflicts

PR #181, opened 2021-06-01
feat(theme-default): improve SidebarButton a11y

Mister-Hope:


packages/@vuepress/theme-default/src/client/components/ToggleSidebarButton.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <div class="toggle-sidebar-button" @click="$emit('toggle')">
+  <div
+    class="toggle-sidebar-button"
+    title="Toggle Sidebar"
```
meteorlxy, line 4: Then the text should be part of the i18n config.

This kind of a11y attrs may increase the i18n fields a lot 🤔

meteorlxy:
Closing as no further progress here

PR #189, opened 2021-06-07
fix(plugin-search): disable default form action on Enter key

KarolisBreidokas:
By default form element on Enter key press would try to send form request to current page, essentially reloading the page.


meteorlxy:
Thanks for catching that.

Could be introduced by #165.

cc @Mister-Hope

PR #208, opened 2021-06-12
fix(bundler-webpack): fix devServer issues

Mister-Hope:
See:

https://github.com/webpack/webpack-dev-server/blob/68e2f0692b8abfee8352a79948592249dfac752f/lib/Server.js#L117
https://github.com/webpack/webpack-dev-server/blob/68e2f0692b8abfee8352a79948592249dfac752f/lib/Server.js#L315
https://github.com/webpack/webpack-dev-server/blob/68e2f0692b8abfee8352a79948592249dfac752f/examples/util.js#L76-L79

The `beforeDevServer` and `afterDevServer` is not just unworkable, but throwing errors now.

### DIFF
--- a/packages/@vuepress/bundler-webpack/src/dev/createDevServerConfig.ts
+++ b/packages/@vuepress/bundler-webpack/src/dev/createDevServerConfig.ts
@@ -28,17 +28,17 @@ export const createDevServerConfig = (
     },
     host: app.options.host,
     hot: true,
-    onAfterSetupMiddleware: ({ app }, server) => {
+    onAfterSetupMiddleware: ({ app: expressApp }, server) => {
       // plugin hook: afterDevServer
-      options.afterDevServer?.(app, server)
+      options.afterDevServer?.(expressApp, server)
     },
-    onBeforeSetupMiddleware: ({ app }, server) => {
+    onBeforeSetupMiddleware: ({ app: expressApp }, server) => {
       // use trailing slash middleware to support vuepress routing in dev-server
       // it will be handled by most of the deployment platforms
-      app.use(trailingSlashMiddleware)
+      expressApp.use(trailingSlashMiddleware)
 
       // plugin hook: beforeDevServer
-      options.beforeDevServer?.(app, server)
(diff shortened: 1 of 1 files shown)


packages/@vuepress/bundler-webpack/src/dev/createDevServerConfig.ts
```diff
@@ -28,11 +28,11 @@ export const createDevServerConfig = (
     },
     host: app.options.host,
     hot: true,
-    onAfterSetupMiddleware: (expressApp, server) => {
+    onAfterSetupMiddleware: ({ app: expressApp }, server) => {
```
meteorlxy, line 31: Then the `server` param could be removed now. And the types should be updated.

meteorlxy (changes requested):

packages/@vuepress/bundler-webpack/src/dev/createDevServerConfig.ts
Mister-Hope: I will fix that tomorrow morning

packages/@vuepress/bundler-webpack/src/dev/createDevServerConfig.ts
```diff
@@ -28,11 +28,11 @@ export const createDevServerConfig = (
...
       // plugin hook: afterDevServer
       options.afterDevServer?.(expressApp, server)
     },
-    onBeforeSetupMiddleware: (expressApp, server) => {
+    onBeforeSetupMiddleware: ({ app: expressApp }, server) => {
       // use trailing slash middleware to support vuepress routing in dev-server
       // it will be handled by most of the deployment platforms
       expressApp.use(trailingSlashMiddleware)
```
meteorlxy, line 38: Why this line does not throw error?

Mister-Hope: This is the expect express app instance which webpack dev server use.

Working in my local env.

PR #218, opened 2021-06-14
feat(theme-default): dynamically change theme-color with dark mode (close #215)

### LINKED ISSUE #215: [Feature request] Dynamically theme-color change with dark mode
### Description

Since Apple introduced new Safari design in WWDC 2021, personally, I think that changing theme-color with dark mode dynamically would have a nicer look for upcoming Safari 15.x (maybe improve other browsers too). Or in another word the current theme-color is too bright for the dark mode, compared in the attached video.

https://user-images.githubusercontent.com/15633984/121815457-412dec00-cca9-11eb-8458-fede21b65966.mov

### Proposed Solution

Not experienced in Vue but I think something like this would did help:

```
// ToggleDarkButton.vue

….

const changeThemeColor = (): void => {
  const themeColorEl = window?.document.querySelector(
    'meta[name="theme-color"]'
  )

  themeColorEl?.setAttribute('content', !isDark.value ? '#3eaf7c' : '#22272e')
}

…

const setDark = (): void => {
  setDarkClass()
  changeThemeColor()
}

onMounted(() => {
  …

  watch(isDark, setDark, { immediate: true })
})
```

If it fine I could pull a request, thanks for reviewing.

Unbinilium:
### Description

Linked with this feature request issue #215 

### Note

For better user experience, the browser should get color hex before the page loaded, but it seems *vuepress-next* is not currently supported duplicated meta tags like this:

```ts
['meta', { name: 'theme-color', content: '#3eaf7c', media: '(prefers-color-scheme: light)' }],
['meta', { name: 'theme-color', content: '#22272e', media: '(prefers-color-scheme: dark)' }]
```

It means the `theme-color` was changed after the script executes if system side is dark mode by default, because I could not have different color scheme in duplicated meta tags currently, and due to some reason that `#3eaf7c` seems too bright that Safari will not use it for address bar color if system side is dark mode by default (darker would works).

Also it may not be the best or proper workaround, some further commit may should be added.


packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
```diff
@@ -49,12 +49,27 @@ const setDarkClass = (): void => {
   htmlEl?.classList.toggle('dark', isDark.value)
 }
 
+const updateThemeColor = (): void => {
+  const themeColorEl = window?.document.querySelector(
+    'meta[name="theme-color"]'
+  )
+  const newThemeColor = getComputedStyle(
```
meteorlxy, line 56: Should explicitly use `window.getComputedStyle`

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
```diff
@@ -49,12 +49,27 @@ const setDarkClass = (): void => {
   htmlEl?.classList.toggle('dark', isDark.value)
 }
 
+const updateThemeColor = (): void => {
+  const themeColorEl = window?.document.querySelector(
+    'meta[name="theme-color"]'
+  )
```
meteorlxy, line 55: If we are going to support mulptiple `meta[name="theme-color"]`, the code here should be updated, too.

meteorlxy (changes requested):

meteorlxy:
We are using following snippet to dedupe `<meta>` tag:

https://github.com/vuepress/vuepress-next/blob/33c9a2145eeaf53429f2ae784e393074d1a9edd5/packages/%40vuepress/shared/src/utils/resolveHeadIdentifier.ts#L11-L14

But I think that supporting multiple `<meta name="theme-color">` tags should be another issue.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
```diff
@@ -49,12 +49,29 @@ const setDarkClass = (): void => {
   htmlEl?.classList.toggle('dark', isDark.value)
 }
 
+const updateThemeColor = (): void => {
+  const themeColorEl = window?.document.querySelectorAll('meta[name="theme-color"]')
+  const newThemeColor = window.getComputedStyle(
```
Unbinilium, line 54:
```suggestion
  const newThemeColor = window.getComputedStyle(
```

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
Unbinilium: https://github.com/vuepress/vuepress-next/blob/b0a9815c9eeaa2ba623cc1a153139e85397d450e/packages/%40vuepress/theme-default/src/client/components/ToggleDarkButton.vue#L54

Got resolved, thanks.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
Unbinilium: https://github.com/vuepress/vuepress-next/blob/b0a9815c9eeaa2ba623cc1a153139e85397d450e/packages/%40vuepress/theme-default/src/client/components/ToggleDarkButton.vue#L52-L62

Now is able to change color for all duplicated 'theme-color' meta tags. I just simply change both light and dark specialized meta tags, because if only change the 'theme-color' meta tag based on current system side color mode would let it not working in some situation:

- system side color mode is currently **light**, if only change the 'theme-color' meta tag matched with`(prefers-color-scheme: light)` to **dark**, then change system side color mode to **dark**, 'theme-color' meta tag matched with`(prefers-color-scheme: light)` is still **dark**, then switch auto enabled dark mode  to light mode by the website's button and change system side color mode back to **light**, the `isDark.value` is updated but not changed, so 'theme-color' meta tag matched with`(prefers-color-scheme: light)` is still in **dark**.

To solve this would be little complicated, personally I think change them both would be better.

Unbinilium:
> We are using following snippet to dedupe `<meta>` tag:
> 
> https://github.com/vuepress/vuepress-next/blob/33c9a2145eeaf53429f2ae784e393074d1a9edd5/packages/%40vuepress/shared/src/utils/resolveHeadIdentifier.ts#L11-L14
> 
> But I think that supporting multiple `<meta name="theme-color">` tags should be another issue.

https://github.com/vuepress/vuepress-next/blob/b0a9815c9eeaa2ba623cc1a153139e85397d450e/packages/%40vuepress/shared/src/utils/resolveHeadIdentifier.ts#L11-L14

It seems little rude to allow duplicated meta name 'theme-color' in this way, but it simple. Or still avoid duplicated  'theme-color', but keeps if they has a different hash.

packages/@vuepress/shared/src/utils/resolveHeadIdentifier.ts
```diff
@@ -8,8 +8,8 @@ export const resolveHeadIdentifier = ([
   attrs,
   content,
 ]: HeadConfig): string => {
-  // avoid duplicated `<meta>` with same `name`
-  if (tag === 'meta' && attrs.name) {
+  // avoid duplicated `<meta>` with same `name` except the name `theme-color`
+  if (tag === 'meta' && attrs.name && attrs.name !== 'theme-color') {
```
meteorlxy, line 12: I'd say this is another issue, and we'd better not change this in current PR.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
meteorlxy, line 54: Consistently use `window?.`

Or skip the following code when `themeColorEl` is empty

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
```diff
@@ -49,12 +49,29 @@ const setDarkClass = (): void => {
...
+  )?.trim()
+  themeColorEl?.forEach((mata) => {
+    mata.setAttribute('content', newThemeColor)
+  })
+}
+
+const setDarkMode = (): void => {
+  setDarkClass()
+  updateThemeColor()
+}
```
meteorlxy, line 67: Let's make the these methods name consistent. I think we can use `updateXXX`.

meteorlxy (changes requested):
There's another problem.

Meta tags set in user config will be injected to client with `useSiteData`. So we also have to update the sitedata ref, or the tag will be reset once navigating to another page.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
Unbinilium: https://github.com/vuepress/vuepress-next/blob/5399ebc21cb05f3495e2b9fb12b620ce40ee28dc/packages/%40vuepress/theme-default/src/client/components/ToggleDarkButton.vue#L69-L72

Ok, this seems would be fine.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
Unbinilium: https://github.com/vuepress/vuepress-next/blob/b84e5dbdf766fe41834d215696f0406647b7cd20/packages/%40vuepress/theme-default/src/client/components/ToggleDarkButton.vue#L52-L67

Thanks for reviewing again, I prefer skip the following code when `themeColorEl` or `newThemeColor` is empty (not very sure the additional check for `newThemeColor` is necessary)

packages/@vuepress/shared/src/utils/resolveHeadIdentifier.ts
Unbinilium: Okey, reverted in https://github.com/vuepress/vuepress-next/pull/218/commits/bf81653b01df39100bbd74f5c725be402c3a2ca8

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
```diff
@@ -44,17 +44,39 @@ const toggleDark = (): void => {
...
+      .trim()
+    if (newThemeColor) {
+      themeColorEl.forEach((mata) => {
+        mata.setAttribute('content', newThemeColor)
+      })
+    }
+  }
+}
+
+const setDarkMode = (): void => {
```
meteorlxy, line 69: `setAppearance` or `setThemeMode` could be better

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
```diff
@@ -44,17 +44,39 @@ const toggleDark = (): void => {
...
 
+const updateThemeColor = (): void => {
+  const themeColorEl = window?.document.querySelectorAll(
+    'meta[name="theme-color"]'
+  )
+  if (themeColorEl) {
+    const newThemeColor = window
+      .getComputedStyle(window.document.documentElement)
+      ?.getPropertyValue(isDark.value ? '--c-bg' : '--c-brand')
+      .trim()
```
meteorlxy, line 60: - The question mark in `?.getPropertyValue` seems useless.
- Is the ending `trim()` necessary?

docs/.vuepress/config.ts
```diff
@@ -47,7 +47,22 @@ export default defineUserConfig<DefaultThemeOptions>({
...
+      },
+    ],
+    [
+      'meta',
+      {
+        name: 'theme-color',
+        content: '#22272e',
+        media: '(prefers-color-scheme: dark)',
+      },
+    ],
```
meteorlxy, line 65: Let's revert this change, too. "Multiple theme-color tag" is not coupled with current PR.

meteorlxy (changes requested):

meteorlxy:
The problem about site data could be handled in this PR. If you are not familiar with that, I'll update it later.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
Unbinilium: 1. I think in some case `getComputedStyle()` may return an empty result, so I added a question mark. It did superfluous because if an empty result has been returned, the website must be broken.

2. Without the `trim()`, `newThemeColor` would always got 1 more space the front of color hex, though the space has effects to the results, but seeing that change `content="#xxxxxx"` -> `content=" #xxxxxx"` would let me feel uncomfortable.

I’m sorry that I'm not familiar with the actual principles for the code style. If u still think it is unnecessary, I can change them later.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
Unbinilium: Done in https://github.com/vuepress/vuepress-next/pull/218/commits/c964abcf35922bbee773777fef203c90e965d705

docs/.vuepress/config.ts
Unbinilium: Okey, reverted in https://github.com/vuepress/vuepress-next/pull/218/commits/3dc1094fc698bb9e8758abd50bd0e3433c4df7fa

Unbinilium:
> The problem about site data could be handled in this PR. If you are not familiar with that, I'll update it later.

I'm not familiar with that and now I may not have enough time to learn and handle that correctly during my term examination (maybe later I cloud have a try). You can also update it, thanks!

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
meteorlxy: 1. Check the MDN docs: https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle
  `getComputedStyle` will always return a live CSSStyleDeclaration object.

2. I didn't see any extra space in chrome. Does it only happen in Safari?

Unbinilium: 1. Thanks, I removed the question mark in commit https://github.com/vuepress/vuepress-next/pull/218/commits/42b3ce7c994d06fe174887ec6ab86f6fe5fd09df

2. In Safari Version 15.0 (17612.1.15.1.14), it would always have an extra space on development (`docs:dev`) but not happen on release. 

Screen Shot 2021-06-15 at 21 49 46

meteorlxy: Got it. It only happens in dev because the CSS is not minified. Let's keep the `trim()` for now.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
```diff
@@ -44,17 +44,39 @@ const toggleDark = (): void => {
...
+const updateDarkClass = (): void => {
   const htmlEl = window?.document.querySelector('html')
   htmlEl?.classList.toggle('dark', isDark.value)
 }
 
+const updateThemeColor = (): void => {
+  const themeColorEl = window?.document.querySelectorAll(
+    'meta[name="theme-color"]'
+  )
+  if (themeColorEl) {
```
meteorlxy, line 56: Let's return early to avoid nested `if`, e.g. `if (!xxxx) return`

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
```diff
@@ -44,17 +44,39 @@ const toggleDark = (): void => {
...
+    const newThemeColor = window
+      .getComputedStyle(window.document.documentElement)
+      .getPropertyValue(isDark.value ? '--c-bg' : '--c-brand')
+      .trim()
+    if (newThemeColor) {
+      themeColorEl.forEach((mata) => {
```
meteorlxy, line 62: Should be a typo `mata` -> `meta`.

meteorlxy (changes requested):
Looks much better.

packages/@vuepress/theme-default/src/client/components/ToggleDarkButton.vue
Unbinilium: Thanks, fixed in commit https://github.com/vuepress/vuepress-next/pull/218/commits/644131a05d330d5588e23ea27a13f59c1452681d

meteorlxy:
Another point:

Currently in this PR, we are using `--c-bg` and `--c-brand` as the theme color. But users could set different theme color in `head` config

Unbinilium:
> Another point:
> 
> Currently in this PR, we are using `--c-bg` and `--c-brand` as the theme color. But users could set different theme color in `head` config

Do you think this workaround would be better? The theme-color now entirely depends on the user's configuration.

```ts
let themeColorLight: string
let themeColorDark: string
const updateThemeColor = (): void => {
  const themeColorEl = window?.document.querySelectorAll('meta[name="theme-color"]')
  if (!themeColorEl) return

  // get current system appearance mode by `prefers-color-scheme`
  const systemAppearanceMode = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light'

  // iterate `theme-color` meta tags
  themeColorEl.forEach((meta) => {
    // get `media` attribute from this meta tag
    const mediaAttr = meta.getAttribute('media')

    // initial themeColorLight or themeColorDark if it is uninitiated
    if (!themeColorLight && mediaAttr?.includes('light')) {
      themeColorLight = meta.getAttribute('content')!
    }
    if (!themeColorDark && mediaAttr?.includes('dark')) {
      themeColorDark = meta.getAttribute('content')!
    }
    
    // if themeColorLight and themeColorDark are initiated, change `content` attribute if the meta `media` attribute matches current system appearance mode
    if (themeColorLight && themeColorLight && mediaAttr?.includes(systemAppearanceMode)) {
      meta.setAttribute('content', isDark.value ? themeColorDark : themeColorLight)
    }
  })
}
```

meteorlxy:
Currently, we could use user config as the light mode theme-color, and use `--c-bg` as the dark mode theme-color implicitly.

Also, the `themeConfig.themeColorLight / themeColorDark` could be an alternative.

---

I've updated this PR with a force push to resolve some conflicts. You could help to check if it works as expected when you have time.

Unbinilium:
Thanks, it works as expected after fix a logic typo in commit https://github.com/vuepress/vuepress-next/pull/218/commits/5bd549845d8758c668b99d5cedace284aee96468. 

For better dark mode `theme-color` support, some new standard like multiple `theme-color` tag would be needed.

https://user-images.githubusercontent.com/15633984/122413967-5cef0600-cfb9-11eb-9f55-4bad71800d6c.mov

Unbinilium:
Considering that simply changing the meta tag of `theme-color` in `siteData` would cause `themeColorHeadContent` not find the original theme-color after repeatedly switching dark mode, and adding a temporary variable to store the original color would be an inelegant way.

I prefer to do this with a new option in `themeConfig`, which changes the value of the content of the `theme-color` meta tag in `siteData` and current value in `<head>` via the value of the `themeColorLight` / `themeColorDark` options in `themeConfig`, and a boolean variable called `themeColor` to control whether or not enable these options.

I hope my implementation in the code is correct, I still have a lot to learn before I can write quality code.

meteorlxy:
Closing as it's stale for a long time. Thanks for all the efforts anyway

meteorlxy:
Closing as it's stale for a long time. Thanks for all the efforts anyway

PR #220, opened 2021-06-15
build(shared): avoid publishing tsbuildinfo

nsznsznjsz:
By default with our config, typescript will generate a `.tsbuildinfo` file in the parent directory of the `outDir`. However, the parent of `./lib/esm` is `./lib` which is the publish dir, we have to explicitly specify path of `.tsbuildinfo`.

This will avoid publish `.tsbuildinfo` file, which has a size of 32kb, the `lib` folder will shrink from 65.1kb to 32.0kb.


Mister-Hope:
That's a nice catch!

PR #238, opened 2021-06-26
fix(theme-default): fix nav glitch at exactly 719px screen width

activefog:
The default theme contains certain styles and logic that execute at a media query breakpoint of 719px screen width. However, there is some overlap of styling and behavior of the nav bar / sidebar at exactly 719px (i.e., behavior is as expected at 718px and 720px, but not 719px). These commits alleviate those issues.


meteorlxy:
Nice catch! Thanks!

BTW, it would be better to provide before/after screenshots when making a UI change.

activefog:
> Nice catch! Thanks!
> 
> BTW, it would be better to provide before/after screenshots when making a UI change.

Thank you for this feedback -- will provide screenshots next time!

PR #239, opened 2021-06-28
docs: fix typo in zh translation

wgytcdx:
选项 ——》 选择


meteorlxy:
Thanks for catching that!

PR #240, opened 2021-06-28
docs(plugin-docsearch): fix wrong dependency name in config example

sparkcyf:
The name `@vuepress/docsearch` listed in the document of the plugin itself was wrong. The name of the plugin in config.ts should be `@vuepress/plugin-docsearch`.

Ref: https://github.com/vuepress/vuepress-next/blob/5fc351d88f9c8e9f6c6af18674aebfd6aafd97f9/docs/.vuepress/config.ts#L155


meteorlxy:
Thanks, but it's legal usage:

https://v2.vuepress.vuejs.org/guide/plugin.html

PR #245, opened 2021-06-29
feat(bundler-vite): avoid global constants being replaced by vite (close #244)

### LINKED ISSUE #244: [Bug report] Defined variables being replaced by vite
### Description

Related: #143, #214

As [warned in our document](https://v2.vuepress.vuejs.org/reference/bundler/vite.html#cautions), this is a known bug. The defined variables are replaced by static strings during the build process when using vite as bundler.

However, this should be a technically possible problem and we can see that this issue has been solved on the vitepress.

https://github.com/vuejs/vitepress/blob/d971dd9996f254c800de06637515d88964779043/src/node/markdownToVue.ts#L49-L63

### Steps to reproduce

- Reproduction link / repo: <https://github.com/nsznsznjsz/vuepress-next-repro-env-variable>

### Expected behavior

Should not throwing an error when writing special words.

### Screenshots

image

### Environment info

- Browser:
- Output of `vuepress info`:

```bash
# Paste output of `vuepress info` here
```

nsznsznjsz:
## Description

Close #244 

This PR provide a solution to resolve the issue by adding a vite plugin. 

### ~~Why do we need to write the html file twice?~~

- ~~This is a special bug for vite, and we only have bundler package that are dependent on vite.~~
- ~~The first time is written in the `@vueprees/core`, by this time the bundler had not yet started working.~~

~~So we have to rewrite it in `bundler-vite`.~~

Update: we are replacing constants by plugin before `@vitejs/plugin-vue` in memory now.


nsznsznjsz:
Is there any way to test it here?

meteorlxy:
Users may use those varibales in a `<script>` tag intentionally in a markdown file, then they'll also be replaced

meteorlxy:
There might be a better way: create a vite plugin to pre-handle `.vue` files before `@vitejs/plugin-vue`

nsznsznjsz:
Got it! I will update it late today.

nsznsznjsz:
I have updated the pr, change it to an vite plugin to avoid the problem of rewriting, and should not replacing constants in hoisted elements.

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
```diff
@@ -0,0 +1,53 @@
...
+          `\\b(${define
+            .map((key) => key.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&'))
+            .join('|')})`,
+          'g'
+        )
+      }
+    },
+    transform(code, id) {
+      if (fileRegex.test(id)) {
+        for (const match of code.matchAll(templateRegex)) {
```
meteorlxy, line 26: `@vue/compiler-sfc` will only recognize one template block. So let's remove the `g` flag and remove the for loop

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
```diff
@@ -0,0 +1,53 @@
...
+              defineRegex,
+              (_) => `${_[0]}<wbr/>${_.slice(1)}`
+            )
+          }
+
+          const index = match.index!
+          code =
+            code.substring(0, index) +
+            html +
+            code.substring(index + html.length)
```
meteorlxy, line 46: I'm afraid that the index would be wrong in the next loop if you change the `code` this way.

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
```diff
@@ -0,0 +1,53 @@
+import type { Plugin } from 'vite'
+
+const fileRegex = /\.html\.vue$/
+const templateRegex = /<template>[\s\S]*<\/template>/g
```
meteorlxy, line 4: Let's use `Regexp` as the suffix.

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
```diff
@@ -0,0 +1,53 @@
...
+export const createConstantReplacementPlugin = (): Plugin => {
+  let defineRegex: RegExp | null = null
+
+  return {
+    name: 'vuepress:constant-replacement',
+    enforce: 'pre',
+    apply: 'build',
+    configResolved(resolvedConfig) {
+      if (resolvedConfig.define) {
+        const define = Reflect.ownKeys(resolvedConfig.define) as string[]
```
meteorlxy, line 15: Is it necessary to use `Reflect.ownKeys` ?

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
```diff
@@ -0,0 +1,53 @@
...
+          // avoid env variables being replaced by vite
+          html = html
+            .replace(/\bimport\.meta/g, 'import.<wbr/>meta')
+            .replace(/\bprocess\.env/g, 'process.<wbr/>env')
+
+          // also avoid replacing vite user defines
+          if (defineRegex) {
+            html = html.replace(
+              defineRegex,
+              (_) => `${_[0]}<wbr/>${_.slice(1)}`
```
meteorlxy, line 38: Only use underscore for unused arguments

packages/@vuepress/bundler-vite/src/plugin/createPlugin.ts
```diff
@@ -16,6 +17,7 @@ export const createPlugin = ({
   isServer: boolean
   isBuild: boolean
 }): Plugin[] => [
+  createConstantReplacementPlugin(),
   createVuePlugin(options.vuePluginOptions),
   createWorkaroundPlugin(),
```
meteorlxy, line 22: Could you help to switch the position of `createVuePlugin` and `createWorkaroundPlugin` :rofl:

packages/@vuepress/bundler-vite/src/plugin/createPlugin.ts
```diff
@@ -4,6 +4,7 @@ import type { App } from '@vuepress/core'
 import type { ViteBundlerOptions } from '../types'
 import { createVuepressPlugin } from './createVuepressPlugin'
 import { createWorkaroundPlugin } from './createWorkaroundPlugin'
+import { createConstantReplacementPlugin } from './createConstantReplacementPlugin'
```
meteorlxy, line 7: Import order should follow alphabet order (we should set up a eslint config for that...)

meteorlxy (changes requested):

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
nsznsznjsz: Can be replaced by `Object.keys` if you think that we need. 

`Reflect` is a built-in object, introduced by es6, and it was supported since [node.js v6.0.0](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect#browser_compatibility)

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
nsznsznjsz: I did not fully understand what you mean, each file will only be transformed once. Will this problem still occur after 27b9f02 that I had remove the loop.

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
nsznsznjsz: Uh uh, `Reflect.ownKeys` also returns non-enumerable prop names or symbols, although this should not be allowed in user define, but we should catch it.

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
meteorlxy: Hmmm, maybe my previous comment is not fully correct.

The `html.length` here is already different from the original `match[0].length`

I wonder if you could get correct result here. 🤔

nsznsznjsz: Fixed! my typo here.

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
meteorlxy: - Vite is using spread operator for user defined `define`, which won't allow non-enumerable properties.
- `Object.keys()` has better readability

nsznsznjsz: Changed on ef94d6a

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
meteorlxy: As for my first comment:

The `code` length is already changed, then the `match.index` in next loop would be incorrect.

Of course it won't happen after removing the loop.

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
```diff
@@ -0,0 +1,54 @@
...
+          // avoid env variables being replaced by vite
+          html = html
+            .replace(/\bimport\.meta/g, 'import.<wbr/>meta')
+            .replace(/\bprocess\.env/g, 'process.<wbr/>env')
+
+          // also avoid replacing vite user defines
+          if (defineRegex) {
+            html = html.replace(
+              defineRegex,
+              (a) => `${a[0]}<wbr/>${a.slice(1)}`
```
meteorlxy, line 39: 😅 And don't use `a` as arg name...

`str` could be better

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
```diff
@@ -0,0 +1,54 @@
...
+        defineRegex = new RegExp(
+          `\\b(${define
+            .map((key) => key.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&'))
+            .join('|')})`,
+          'g'
+        )
+      }
+    },
+    transform(code, id) {
+      if (fileRegexp.test(id)) {
```
meteorlxy, line 25: Return early to avoid nested `if` block.

nsznsznjsz:
Done!

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
nsznsznjsz: Incredible! I never realized it was a bug, thanks for your careful consideration and points out ❤️

packages/@vuepress/bundler-vite/src/plugin/createConstantReplacementPlugin.ts
nsznsznjsz: Changed to `define`

meteorlxy:
When those constants appear in page headers, they will be extracted to page data. So we also need to replace them in page data files.

Just rebase the branch and made some modifications. @nsznsznjsz You can help to check if there are anything wrong.

meteorlxy (approved):

nsznsznjsz:
@meteorlxy thanks for your review

PR #259, opened 2021-07-04
fix(plugin-active-header-links): update active header link selector

Danktuary:
Since the SidebarChild's template got updated from `.sidebar-link` to `.sidebar-item` in ea7c4bb, the default value for `headerLinkSelector` on the active-header-links plugin doesn't exist anymore (at least not on the default theme).


meteorlxy:
Oops. Thanks for catching that!

PR #274, opened 2021-07-10
feat(theme-default): improve SidebarButton a11y

Mister-Hope:


meteorlxy (changes requested):
Same as #275.

BTW, you could help to migrate this component to `<script setup>`, too

Mister-Hope:
I will just leave this commit till you merge #275, because I think it will may get conflict when that pr is merged. I will continue after you merge #275

PR #275, opened 2021-07-10
feat(theme-default): add `toggleDarkMode` locale option

Mister-Hope:


Mister-Hope:
This should fix #270

packages/@vuepress/theme-default/src/shared/options.ts
```diff
@@ -79,6 +79,11 @@ export interface DefaultThemeLocaleData extends LocaleData {
    */
   darkMode?: boolean
 
+  /**
+   * A11y text for darkmode toggle button
+   */
+  darkmodeText?: string
```
meteorlxy, line 85: Should use the same camelCase as the `darkMode` option

packages/@vuepress/theme-default/src/client/components/ToggleDarkModeButton.vue
```diff
@@ -42,10 +42,16 @@
...
+import { computed } from 'vue'
+import { useDarkMode, useThemeLocaleData } from '../composables'
 
 const isDarkMode = useDarkMode()
 const toggleDarkMode = (): void => {
   isDarkMode.value = !isDarkMode.value
 }
+
+const title = computed(() => {
+  const themeLocale = useThemeLocaleData()
```
meteorlxy, line 54: Put the `use` function outside computed

meteorlxy: In fact, we may not need a computed value here. We should set the default value in `assignDefaultLocaleOptions`

meteorlxy (changes requested):
And the docs should be updated, too.

Mister-Hope:
Just find out I format the rest of the docs when saving, but I think that's fine, if you refuse it I will revert and recommit the docs changes.

packages/@vuepress/theme-default/src/client/components/ToggleDarkModeButton.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <button class="toggle-dark-button" @click="toggleDarkMode">
+  <button
+    class="toggle-dark-button"
+    :title="title"
+    @click="themeLocale.darkModeText"
```
meteorlxy, line 5: ... I think you should test PR locally first 😅

Mister-Hope:
My fault, I tested before opening it, but I am lazy testing it after a refactor 🌚

PR #276, opened 2021-07-10
feat(markdown): support `code.lineNumbers` to be set to number (close #231)

### LINKED ISSUE #231: [Feature request] Automatically disable code line numbers for code blocks with <= 3 lines
Small code block don't benefit from line numbers and look neater without them.

Mister-Hope:


docs/reference/config.md
```diff
@@ -333,13 +333,14 @@ You should not configure it unless you understand what it is for.
 
 #### markdown.code.lineNumbers
 
-- Type: `boolean`
+- Type: `boolean | number`
 
-- Default: `true`
+- Default: `4`
```
meteorlxy, line 338: I think we should not change the default value.

docs/reference/config.md
```diff
@@ -333,13 +333,14 @@ You should not configure it unless you understand what it is for.
...
 
 - Details:
 
-  Enable code line numbers or not.
```
meteorlxy, line 342: The first line is a general description of the option, and should not be removed.

docs/reference/config.md
```diff
@@ -333,13 +333,14 @@ You should not configure it unless you understand what it is for.
...
+  - `boolean` means enable code line numbers or not.
+  - `number` means the threshold line number enabling it
```
meteorlxy, line 343: Should refine the decription of `number` value. Consider providing an example.

meteorlxy (changes requested):

PR #282, opened 2021-07-11
docs: format deployment & update actions-gh-pages

YunYouJun:
- format deployment.md with [markdownlint](https://github.com/DavidAnson/markdownlint)
  - Add indent for list
- use [actions-gh-pages](https://github.com/peaceiris/actions-gh-pages) instead of [ghaction-github-pages](https://github.com/crazy-max/ghaction-github-pages)
  - The former is more popular, and updates are active.
  - More detailed documentation.
  - Its name is more friendly, and it is deployed to gh-pages by default.

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -19,119 +19,124 @@ The following guides are based on some shared assumptions:
 
 1. Set the correct [base](../reference/config.md#base) config.
 
-    If you are deploying to `https://<USERNAME>.github.io/`, you can omit this step as `base` defaults to `"/"`.
+   If you are deploying to `https://<USERNAME>.github.io/`, you can omit this step as `base` defaults to `"/"`.
 
-    If you are deploying to `https://<USERNAME>.github.io/<REPO>/`, for example your repository is at `https://github.com/<USERNAME>/<REPO>`, then set `base` to `"/<REPO>/"`.
+   If you are deploying to `https://<USERNAME>.github.io/<REPO>/`, for example your repository is at `https://github.com/<USERNAME>/<REPO>`, then set `base` to `"/<REPO>/"`.
 
 2. Choose your preferred CI tools. Here we take [GitHub Actions](https://github.com/features/actions) as an example.
 
-    Create `.github/workflows/docs.yml` to set up the workflow.
+   Create `.github/workflows/docs.yml` to set up the workflow.
 
-::: details Click to expand sample config
-```yaml
-name: docs
+   ::: details Click to expand sample config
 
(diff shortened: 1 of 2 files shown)


meteorlxy:
Don't format like that. The indentation is avoided intentionally

meteorlxy:
Closing as it's not necessary to replace a project which is working properly. And it's only an example, users can choose other projects themselves

YunYouJun:
> Don't format like that. The indentation is avoided intentionally

I am not simply using markdownlint for automatic formatting.

I noticed that the indentation after the list is inconsistent, some are indented, and some are not.

image

After the changes are made, the uniformity of the style can be maintained, and no problems caused by this have been found so far.

image

YunYouJun:
> Closing as it's not necessary to replace a project which is working properly. And it's only an example, users can choose other projects themselves

I think that when the functions of the projects are almost the same, it is necessary to use projects with more detailed documentation and more active maintenance as examples. Users can more easily find some potential solutions from the project.

Although the choice of items is free, users will follow the document examples more to use, and will not think of another good choice.

YunYouJun:
Of course, these are my personal opinions. If I can't persuade you, just leave it here.

Mister-Hope:
I agree with you maybe we can add a linter (some mistaken trailling white space or missing some white space is bothering me too), but the action is just personal taste.

E.g.: for me I am using https://github.com/JamesIves/github-pages-deploy-action

YunYouJun:
> I agree with you maybe we can add a linter (some mistaken trailling white space or missing some white space is bothering me too), but the action is just personal taste. If it's working fine, and does work fine in most situations, then it's good to use.
> 
> E.g.: for me I am using https://github.com/JamesIves/github-pages-deploy-action

Yeah, I see, I can separate them if needed.

meteorlxy:
If you take a look at the file history, you might find that we used to have those indents. However, it looks bad on mobile devices.

YunYouJun:
I may understand what you mean.

Do you mean this?

image

image

Is it a better way to customize the code block style in `<li></li>` for the mobile?

Mister-Hope:
I think we can use tools (whichi won't be much strict) like prettier to format markdown, I just tried, prettier won't force indent that list.

Just like [here](https://github.com/vuepress/vuepress-next/pull/275/commits/471ee26c60939c0bc305231415661fd54dd38f6c#diff-a80a5ac6d641cce4f0b8e2c4c785b7529eed8069c21f79de9466029bd53d3537) in line 61, these issues won't be nice.

PR #288, opened 2021-07-19
fix(theme-default): fix navbar active styles

maomao1996:
[demo](https://71noe.sse.codesandbox.io/guide/page.html)

WeChatdeeaf6281487a3cb310f4ec513407e34


meteorlxy:
Check the docs. Use `activeMatch` for that

PR #289, opened 2021-07-19
feat(theme-default): set dark mode color-scheme

jiazengp:


meteorlxy:
`vars-dark.scss` is for variables only. We should put it to another place 🤔

jiazengp:
It looks like it should be placed in normalize.scss😁

jiazengp:
previous:
Snipaste_2021-07-20_13-41-48
current:
Snipaste_2021-07-20_13-41-33

Issue #283, opened 2021-07-12
[Feature request] themeConfig logo darkmode dependent


hschaefer123:
## Feature request
Allowing themeConfig logo setting for light- and darkMode.

### Description
The themeConfig setting logo should allow to differ between light/darkMode to allow different logo on dark screens.

### Proposed Solution
Maybe just an addtional property like
darkModeLogo: '/dark-logo.svg'
that should be used in dark mode instead of using logo.

This would allow to optimize logos for dark and light mode, especially in cases where logo contains text slogan, that should have a different text color.

Best Regards
Holger

PR #295, opened 2021-07-20
fix(theme-default): use a global isDarkMode ref

Shigma:
This PR uses a global ref to make sure `useDarkMode()` still be of actual use in inherited theme development.

### DIFF
--- a/packages/@vuepress/theme-default/src/client/components/ToggleDarkModeButton.vue
+++ b/packages/@vuepress/theme-default/src/client/components/ToggleDarkModeButton.vue
@@ -48,14 +48,14 @@
 <script setup lang="ts">
 import {
   useDarkMode,
-  useDocumentClass,
+  useHTMLClass,
   usePrefersColorScheme,
   useThemeLocaleData,
 } from '../composables'
 
 const isDarkMode = useDarkMode()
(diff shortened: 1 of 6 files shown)


Mister-Hope:
Why should a plugin require theme files?

Shigma:
> Why should a plugin require theme files?🌚User may use other themes.

I corrected my words. I'm writing my documentation site with a theme extending @vuepress/theme-default, and just wondering how I can get the real-time color scheme.

meteorlxy:
We'd better split the original `useDarkMode` into two different composition func.

- `useDarkMode()`: return the global ref
- `usePrefersColorScheme(useDarkMode())`: watch the media query and update the ref

meteorlxy:
We'd better split the original `useDarkMode` into two different composition func.

- `useDarkMode()`: return the global ref
- `usePrefersColorScheme(useDarkMode())`: watch the media query and update the ref

meteorlxy:
We'd better split the original `useDarkMode` into two different composition func.

- `useDarkMode()`: return the global ref
- `usePrefersColorScheme(useDarkMode())`: watch the media query and update the ref

meteorlxy:
This PR would be a preparation of #283

PR #297, opened 2021-07-20
feat(theme-default): extract LayoutFrame component

Shigma:
This PR extracts a `LayoutFrame` component from `Layout`. This is a component which renders as a layout with navbar and sidebar, without page content. It will be really useful for inherited theme developers to create custom layouts.


meteorlxy:
Closing as it's not necessary to do such refactor

PR #321, opened 2021-07-30
feat(theme-default): enhance color transitions

Afanyiyu:
like #287

PR #325, opened 2021-07-31
feat(theme-default): add css variables for transition

Afanyiyu:
From [this review](https://github.com/vuepress/vuepress-next/pull/321#pullrequestreview-719562123)

### DIFF
--- a/packages/@vuepress/theme-default/src/client/styles/badge.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/badge.scss
@@ -9,7 +9,8 @@
   padding: 0 6px;
   color: var(--c-bg);
   vertical-align: top;
-  transition: $TColor, $TBackgroundColor;
+  transition: color var(--t-color-options),
+    background-color var(--t-color-options);
 
   &.tip {
     background-color: var(--c-badge-tip);
--- a/packages/@vuepress/theme-default/src/client/styles/badge.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/badge.scss
@@ -1,5 +1,3 @@
-@import '_variables';
-
 .badge {
   display: inline-block;
   font-size: 14px;
--- a/packages/@vuepress/theme-default/src/client/styles/badge.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/badge.scss
@@ -7,8 +7,7 @@
   padding: 0 6px;
   color: var(--c-bg);
   vertical-align: top;
-  transition: color var(--t-color-options),
-    background-color var(--t-color-options);
+  transition: color var(--t-color), background-color var(--t-color);
 
   &.tip {
     background-color: var(--c-badge-tip);
--- a/packages/@vuepress/theme-default/src/client/styles/badge.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/badge.scss
@@ -1,3 +1,5 @@
+@import '_variables';
+
 .badge {
   display: inline-block;
   font-size: 14px;
@@ -7,7 +9,7 @@
   padding: 0 6px;
   color: var(--c-bg);
   vertical-align: top;
-  transition: var(--t-p-color), var(--t-p-background-color);
+  transition: $TColor, $TBackgroundColor;
 
   &.tip {
     background-color: var(--c-badge-tip);
--- a/packages/@vuepress/theme-default/src/client/styles/badge.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/badge.scss
@@ -7,9 +7,7 @@
   padding: 0 6px;
   color: var(--c-bg);
   vertical-align: top;
-  transition:
-    background-color ease 0.3s,
-    color ease 0.3s;
+  transition: var(--t-color-all);
 
   &.tip {
     background-color: var(--c-badge-tip);
--- a/packages/@vuepress/theme-default/src/client/styles/badge.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/badge.scss
@@ -7,7 +7,7 @@
   padding: 0 6px;
   color: var(--c-bg);
   vertical-align: top;
-  transition: var(--t-color), var(--t-background-color);
+  transition: var(--t-p-color), var(--t-p-background-color);
 
   &.tip {
     background-color: var(--c-badge-tip);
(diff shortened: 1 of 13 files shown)


Afanyiyu (commented):
https://github.com/vuepress/vuepress-next/pull/325/commits/b340f27aa963c1b9bd7d0ab5438db68bf77195fb

Removed `--t` vars in other plugins. This will cause `transform duration` of these plugins to remain unchanged when user modifies `--t-duration`.

packages/@vuepress/theme-default/src/client/styles/vars.scss
```diff
@@ -65,6 +65,18 @@
...
+  --t-duration: 0.3s;
+  --t-duration-slide: 0.3s;
+  --t-all: all var(--t-duration);
+  --t-opacity: opacity var(--t-duration);
+  --t-color: color var(--t-duration);
+  --t-border-color: border-color var(--t-duration);
+  --t-background-color: background-color var(--t-duration);
+  --t-color-all: var(--t-color), var(--t-border-color),
+    var(--t-background-color);
+  --t-transform: transform var(--t-duration);
```
meteorlxy, line 79: Seems that transition-timing-function is not supported?

Afanyiyu: https://github.com/vuepress/vuepress-next/pull/325/commits/74ff30470f4631ccc0f199f4ad688c62a653cb2b

packages/@vuepress/theme-default/src/client/styles/transitions.scss
```diff
@@ -1,9 +1,9 @@
 .fade-slide-y-enter-active {
-  transition: all 0.3s ease;
+  transition: all var(--t-duration-slide) ease;
 }
 
 .fade-slide-y-leave-active {
-  transition: all 0.3s cubic-bezier(1, 0.5, 0.8, 1);
+  transition: all var(--t-duration-slide) cubic-bezier(1, 0.5, 0.8, 1);
```
meteorlxy, line 6: As the other parts of are not customizable, I think we could just leave the duration constant?

packages/@vuepress/theme-default/src/client/styles/vars.scss
```diff
@@ -65,6 +65,20 @@
   --sidebar-width-mobile: calc(var(--sidebar-width) * 0.82);
   --content-width: 740px;
   --homepage-width: 960px;
+
+  // transition vars
+  --t-duration: 0.3s;
+  --t-duration-slide: 0.3s;
+  --t-timing-function: ease;
```
meteorlxy, line 72: Maybe we could merge them into a single var, because "duration" and "timing function" always appear together in following usage.

meteorlxy (changes requested):

packages/@vuepress/theme-default/src/client/styles/transitions.scss
Afanyiyu: 9c8bf00

packages/@vuepress/theme-default/src/client/styles/vars.scss
Afanyiyu: 822d845

OK. Users can control `transition-duration`, `transition-timing-function` and `transition-delay` in a single variable in this way.

packages/@vuepress/theme-default/src/client/styles/vars.scss
```diff
@@ -65,6 +65,17 @@
...
+  // transition vars
+  --t-options: 0.3s ease;
+  --t-all: all var(--t-options);
+  --t-opacity: opacity var(--t-options);
+  --t-color: color var(--t-options);
+  --t-border-color: border-color var(--t-options);
+  --t-background-color: background-color var(--t-options);
+  --t-color-all: var(--t-color), var(--t-border-color),
+    var(--t-background-color);
+  --t-transform: transform var(--t-options);
```
meteorlxy, line 78: One more thing: there are two different naming patterns

- `--t-${propertyName}`: `--t-all`, `--t-opacity`, `--t-color`
- `--t-${somethingElse}`: `--t-options`, `--t-color-all`

Any good idea to normalize the naming pattern?

meteorlxy: Maybe an extra prefix `--t-p` for property name? 🤔

Afanyiyu: 10fe60f

packages/@vuepress/theme-default/src/client/styles/vars.scss
```diff
@@ -65,6 +65,17 @@
...
+  // transition vars
+  --t-options: 0.3s ease;
+  --t-p-all: all var(--t-options);
+  --t-p-opacity: opacity var(--t-options);
+  --t-p-transform: transform var(--t-options);
+  --t-p-color: color var(--t-options);
+  --t-p-border-color: border-color var(--t-options);
+  --t-p-background-color: background-color var(--t-options);
+  --t-color-all: var(--t-p-color), var(--t-p-border-color),
+    var(--t-p-background-color);
```
meteorlxy, line 78: Well, both @Shigma and me think we might not need so many vars.

IMO, only the `--t-options` is helpful for reusing the transition options. 🤔

Shigma: Although this PR is intended to provide scalability, I believe consistency is also important to a theme.

For theme-default, two transition variables are enough. One for color (color, border-color, background-color, etc) and the other for position (left, right, top, bottom, width, height, transform, etc).

Afanyiyu: 3906192

What naming convention is better to use in `_variables.scss`?

meteorlxy: I think using `color var(--t-color-options)` directly is enough 🤔  No need to add extra scss variables

Afanyiyu: 2908205

packages/@vuepress/theme-default/src/client/styles/badge.scss
```diff
@@ -1,3 +1,5 @@
+@import '_variables';
```
meteorlxy, line 1: Should be removed

meteorlxy (changes requested):

packages/@vuepress/theme-default/src/client/styles/badge.scss
Afanyiyu: 368cb08

meteorlxy (approved):
LGTM. What's your opinion @Shigma ?

PR #365, opened 2021-08-09
feat(theme-default): improve SidebarButton a11y and animate

jiazengp:


packages/@vuepress/theme-default/src/shared/options.ts
```diff
@@ -253,6 +253,11 @@ export interface DefaultThemeLocaleData extends LocaleData {
    */
   openInNewWindow?: string
 
+  /**
+   * A11y text for sidebar toggle button
+   */
+  toggleSidebar?: string
```
meteorlxy, line 259: Should put after `toggleDarkMode`

meteorlxy (changes requested):

packages/@vuepress/theme-default/src/shared/options.ts
jiazengp: All right

meteorlxy (approved):
LGTM

PR #388, opened 2021-08-16
feat(markdown): line numbers from section

jeapostrophe:
Right now, if you extract a section of external code, such as lines 10 to 13, and add line numbers, the numbers are always starting from `1`, rather than from `10`. This changes that so the line numbers are for the actual file.

### DIFF
--- a/packages/@vuepress/markdown/src/plugins/codePlugin/codePlugin.ts
+++ b/packages/@vuepress/markdown/src/plugins/codePlugin/codePlugin.ts
@@ -108,8 +108,15 @@ export const codePlugin: PluginWithOptions<CodePluginOptions> = (
     // generate line numbers
     if (useLineNumbers) {
       // generate line numbers code
+      let lineBase = token.meta?.lineStart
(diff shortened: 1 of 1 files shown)


jeapostrophe:
Obviously this is not backwards compatible, so if you want to, it makes sense to make a configuration option for this. This doesn't seem "plugin"-able, so I didn't try to go that direction.

meteorlxy:
Then I think we should also sync the "highlight lines" syntax to match the exact line numbers. However, as you said, it will introduce some breaking changes, and make the line ranges / numbers inconsistent with other code blocks.

For now I think you'd better fork a `codePlugin` to meet you need.

Mister-Hope:
I agree with @meteorlxy. If you want to match the exact linenumber from file. you would better make a code plugin yourself and change the code block outlook into something like:

```
.. | ..
10 | xxxxx
11 | xxxx
12 | xxxxxxxxxxxxxxxxxxxx
13 | xxxx
.. | ...
```

This is a proper way to make things clearly.

PR #395, opened 2021-08-18
feat(client): allow html title customization (close #338)

### LINKED ISSUE #338: [Feature request] Allow us to Remove site name from page title
I want to remove my site name from every page I don’t want it every where I just want page title

I want to remove the highlighted part in image It will be great if we can do some configuration in config file.

meteorlxy:
Close #338 

This PR allows users to customize the content of `<title>` tag in this way:

```ts
// clientAppEnhance.ts
import { defineClientAppEnhance, resolvePageHeadTitle } from '@vuepress/client'

export default defineClientAppEnhance(() => {
  resolvePageHeadTitle.value = (page, site) => `${page.title} | ${site.title}`
})
```

### DIFF
--- a/packages/@vuepress/client/src/app.ts
+++ b/packages/@vuepress/client/src/app.ts
@@ -104,7 +104,7 @@ export const createVueApp: CreateVueAppFunction = async () => {
   )
   const pageFrontmatter = computed(() => resolvePageFrontmatter(pageData.value))
   const pageHeadTitle = computed(() =>
-    resolvePageHeadTitle(pageData.value, siteLocaleData.value)
+    resolvePageHeadTitle.value(pageData.value, siteLocaleData.value)
   )
   const pageHead = computed(() =>
     resolvePageHead(
--- a/packages/@vuepress/client/src/injections/pageHeadTitle.ts
+++ b/packages/@vuepress/client/src/injections/pageHeadTitle.ts
@@ -1,4 +1,4 @@
-import { inject } from 'vue'
+import { inject, ref } from 'vue'
 import type { ComputedRef, InjectionKey } from 'vue'
 import type { PageData } from './pageData'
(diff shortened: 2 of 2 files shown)


meteorlxy:
cc @Shigma @Mister-Hope @jiazengp

Shigma:
To me this feature is more like a configuration. It will be quite annoying if one has to write lots of different composition functions for every similar configurations.

I suggest that we only provide a simple client API that exposes all configurations.

```ts
// clientAppEnhance.ts
import { defineClientAppEnhance, useClientAppConfig } from '@vuepress/client'

export default defineClientAppEnhance(() => {
  const config = useClientAppConfig() // returns reactive object
  config.resolvePageHeadTitle = (page, site) => `${page.title} | ${site.title}`
})
```

or we can even provide a syntactic sugar for the code above:

```ts
// clientAppEnhance.ts
import { defineClientAppEnhance } from '@vuepress/client'

export default defineClientAppEnhance({
  resolvePageHeadTitle: (page, site) => `${page.title} | ${site.title}`,
})
```

In that case we can easily extend our client app configuration in the future.

meteorlxy:
@Shigma Sounds good to me.

What other config can be put into it (or in the future) ?

Mister-Hope:
My opinion:

I am not coming up any idea what config need to added on the client side, but I think maybe we could extend more config in user config or core.

Some of them already appears in issue or dicussion, and are reasonable.

E.g.:

- Desides whether accessing other pages data sync or async. Small docs or blog sites could benefit from visiting a sync pages array. And large sites can increase speed when asynchronously load each page data.

- add darkmode comparability and its status to core. These will make the srr render result default to dark possible, and help hold the same config and behaviour across plugins and themes.

jiazengp:
So do I

Issue #410, opened 2021-08-27
[Feature request] More navbar slots: navbar-before, navbar-after, navbar-links-before, and navbar-links-after


vmihailenco:
Currently `navbar-before` and `navbar-after` slots are actually inserted before and after navbar **links**. But it would be nice to have some way to extend navbar itself, for example, add another fixed bar above the navbar. So I suggest to:
- rename navbar-before to navbar-links-before, navbar-after to navbar-links-after
- add navbar-before and navbar-after in a similar way navbar-links slots are added now

Alternatively new slots can be named `navbar-top` and `navbar-bottom` since they should be part of the navbar:

```vue
<header ref="navbar" class="navbar">
  <slot name="top" />

  ...

  <slot name="bottom" />
</header>
```

It is possible to achieve the same by extending the theme and I already have done that, but it is kinda messy to have 2 fixed bars and adjust position of other elements on the page twice for each fixed bar. With the proposed solution I could just use the navbar-top slot and change `navbar-height` CSS var.

I can send a PR if this is accepted.

PR #414, opened 2021-08-28
feat(theme-default): add --header-height CSS variable

vmihailenco:
This PR adds `--header-height` CSS variable which is equal to `--navbar-height` by default. When customizing `--header-height: 5.6rem`, the page contents is shifted by 2rem downwards like this

vuepress-header-height

Which then allows to use the space to customize the header, for example, adding another header above/below the existing one.

vp-system-bar

Closes https://github.com/vuepress/vuepress-next/issues/410


meteorlxy:
It's not a common request. You can just override the styles yourself.

PR #474, opened 2021-10-08
fix(theme-default): display code lines as table-cell (close #459)

### LINKED ISSUE #459: [Bug report] Line number and code misalignment
### Description

Line numbers and codes are misaligned on both iOS and macOS Safari, especially when window scaling is performed, seems caused by unwanted line wrap.

### Steps to reproduce

- Reproduction link / repo: [https://v2.vuepress.vuejs.org/reference/cli.html#dev](https://v2.vuepress.vuejs.org/reference/cli.html#dev) 

### Expected behavior

Line numbers and codes are misaligned when your scale down the window width.

### Screenshots 

https://user-images.githubusercontent.com/15633984/134792883-7d9e196c-afb8-4d59-8ab3-44eb368dc719.mov

### Environment info

- Browser: Safari 15.0(17612.2.4.1.2)
- VuePress: v2.0.0-beta.26

ekarad1um:
Fix line number and code line misalignment, close #459.

Verified works on:
- Safari iOS/iPadOS `15.0.1`
- Safari `15.1` macOS
- Chrome `94.0.4606.81` macOS
- Chrome `78.0.3904.96` Android

Screen Shot 2021-10-08 at 17 58 46

Firefox and Edge not tested, I hopefully there’s someone could help for testing.

### DIFF
--- a/packages/@vuepress/theme-default/src/client/styles/code.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/code.scss
@@ -124,6 +124,7 @@ pre[class*='language-'] {
 .theme-default-content {
   pre,
   pre[class*='language-'] {
+    text-size-adjust: 100%;
     line-height: 1.4rem;
     padding: 1.25rem 1.5rem;
     margin: 0.85rem 0;


Unbinilium:
> Not 100% related but nice to have:
> 
> Could you also add a `color-scheme: dark` CSS proprty as well? This would make sure that the code scrollbar is dark in supported browsers.
> 
> For reference see [this StackOverflow question](https://stackoverflow.com/questions/65940522/how-do-i-switch-to-chromes-dark-scrollbar-like-github-does).

I prefer consider this in a separate PR, personally I think the style of the scrollbar should be similar to navBar's but with the color which matches code blocks, simply  add `color-scheme: dark` seems not elegant.

PR #509, opened 2021-11-03
grammar correction

matej-marjanovic:


matej-marjanovic (commented):
Yes, this is better. Thanks.

PR #518, opened 2021-11-07
feat(core): support injecting meta into router

Mister-Hope:


packages/@vuepress/core/src/page/resolvePageData.ts
```diff
@@ -19,14 +19,17 @@ export const resolvePageData = async ({
...
 
   // plugin hook: extendsPageData
   const extendsPageData = await app.pluginApi.hooks.extendsPageData.process(
     page,
     app
   )
-  extendsPageData.forEach((item) => Object.assign(data, item))
+  extendsPageData.forEach((item) =>
+    Object.assign(data, { item, meta: { ...data.meta, ...item.meta } })
+  )
```
Mister-Hope, line 32: Here I deep merge `meta` key to avoid the frontter edits being coverd.

But, I think there are issues with this design. I will open a new issue to explain why there are problems.

Mister-Hope:
I think it's ready for the feature, improvements can make with "frontmatter" when having later discussions

meteorlxy:
Sorry for delay. (Really busy and kind of tired these days 😅 ) I'll pick up this PR soon.

docs/reference/plugin-api.md
```diff
@@ -196,11 +196,15 @@ In client component:
...
+import { useResolveRouteWithRedirect } from '@vuepress/theme-default'
 
 export default {
   setup() {
     const page = usePageData()
-    console.log(page.value.meta) // foobar
+    const route = useResolveRouteWithRedirect('/')
+
+    console.log(page.value.meta.foo) // 'bar'
+    console.log(route.meta.path) // '/'
```
meteorlxy, line 207: As `meta` is already available in the `route.meta`, is it necessary to put `meta` in to page data? 

I think we can make `meta` a property of the `Page` object, and don't need to put it into page data.

What's your opinion? @Mister-Hope

Mister-Hope: Sounds ok to me.

meteorlxy: If we remove it from page data, we could not use `extendsPageData` to modify it. I have two proposals:

1. Let users to modify it manually in `onInitialized` hook:

```ts
onInitialized: (app) => app.pages.forEach(....)
```

In fact, the `extendsPageData` hook could also be removed as we have added `page.data` property, and users can modify it in this way, too. 😅 

2. Rename `extendsPageData` to `extendsPage`:

```ts
extendsPage: (page) => {
  page.data.foo = 'foo'
  page.routeMeta.title = page.title
  // ...
}
```

The only difference is that plugins / themes don't need to traverse the `app.pages` array by themselves.

---

@Mister-Hope

meteorlxy:
I've replaced `extendsPageData` with `extendsPage` in 827a873.

Then this PR could be implemented in a different way.

meteorlxy:
Implemented in de49f14

Issue #544, opened 2021-12-04
[Bug report] default theme NavbarGroup Item was actived falsely.


smart-town:
### Description

When I set navbar like this:   
```js
{
    navbar: [
        { text: 'os', children: [ '/os/', '/os/linux_orders/' ] }
    ]
}
```
Then I can get a  `NavbarGroup` on the page. When I click `NavItem1`(link to `/os/linux_orders/`), then `NavItem1` was actived! But the item(link to `/os/`) was also actived! Both of them are green.(actived).

When I only click `NavItem2`(link to `/os/`), Vuepress works well.

### Steps to reproduce

1. init a vuepress-next project
2. create a folder like this:
    ```shell
    \OS
    |   linux_secure_check.md
    |   README.md
    |
    \---linux_orders
            001cut.md
            README.md
    ```
3. start app, and click the NavItem(linked to `/os/linux_orders/`)
- [Reproduction repo](https://github.com/smart-town/bug_report/tree/vuepress_next)

### Expected behavior

It should only one NavItem was actived when I just choosed one of them.( in NavbarGroup)
### Screenshots
*Both of them are actived:*

![image](https://user-images.githubusercontent.com/16263772/144716306-ad7ce96d-8061-419b-b997-04c7c5bb04e4.png)

*Correct display:*
![image](https://user-images.githubusercontent.com/16263772/144716340-32dcef0b-ed8d-4288-a453-798308d268ef.png)

### Environment info

- Browser: Microsoft Edge 96.0.1054.43
- Output of `vuepress info`:

```bash
# Paste output of `vuepress info` here
System:
    OS: Windows 10 10.0.22000
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-11300H @ 3.10GHz
    Memory: 5.75 GB / 15.79 GB
  Binaries:
    Node: 14.16.1 - D:\javascript\nodejs\node14\node.EXE
    Yarn: Not Found
    npm: 7.21.1 - D:\javascript\nodejs\node14\npm.CMD
  Utilities:
    Git: 2.31.1.
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.22000.120.0), Chromium (96.0.1054.43)
  npmPackages:
    @vuepress/bundler-vite:  2.0.0-beta.27
    @vuepress/bundler-webpack: Not Found
    @vuepress/cli:  2.0.0-beta.27
    @vuepress/client:  2.0.0-beta.27
[...]

PR #545, opened 2021-12-04
:bug: fix issue #544

smart-town:
only add `activeMath` attribute for the url config in navbar


meteorlxy:
`activeMatch` is already supported.

PR #560, opened 2021-12-10
feat(theme-default): add heroImageDark

alirezanet:


alirezanet:
I just realized someone else created similar PR.

PR #562, opened 2021-12-11
docs：@vuepress/plugin-search@next use yarn

Tinywan:


meteorlxy:
Thanks, but we are using `npm i -D` in all plugins docs.

PR #618, opened 2022-01-09
refactor(plugin-active-header-links): improve code structure

Mister-Hope:


packages/@vuepress/plugin-active-header-links/src/client/composables/useActiveHeaderLinks.ts
```diff
@@ -107,13 +107,14 @@ export const useActiveHeaderLinks = ({
   }
 
   const onScroll = debounce(() => setActiveRouteHash(), delay)
+  const listener = (): Promise<void> => onScroll()
```
meteorlxy, line 110: What about normalizing those var names?

onScroll -> setActiveRouteHashDebounce
listener -> onScroll

Or any better advices ?

meteorlxy: This might look better? 🤔 

```ts
const setActiveRouteHash = debounce(() => { ... }, delay)
const onScroll = (): Promise<void> => setActiveRouteHash()
```

Mister-Hope: Yes

meteorlxy (approved):

PR #619, opened 2022-01-09
chore(theme-default): remove unused variable

Mister-Hope:


meteorlxy (approved):
Any suggestions for how to avoid this in vue files? 🤔

PR #621, opened 2022-01-11
refactor(theme-default): extract common code

zh-lx:
extract common code in conditional statements


Mister-Hope:
I am totally disagree with this change, imagine I am building a search plugin where I provide a shortcut `s` for opening the search box.

The current code won't stop opening search box when pressing `s` when focusing this component, but your changes will.

zh-lx:
> I am totally disagree with this change, imagine I am building a search plugin where I provide a shortcut `s` for opening the search box.
> 
> The current code won't stop opening search box when pressing `s` when focusing this component, but your changes will.
> 
> The `<CodeGroup` should only catch those keys which has effect, instead of stoping the whole propagation

Sorry, I didn't think about it properly

PR #634, opened 2022-01-17
fix(bundler-webpack): bump mini-css-extractor

theniceangel:
fix breaking change bug of [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin)


meteorlxy:
Seems that it's already fix by mini-css-extract-plugin itself: https://github.com/webpack-contrib/mini-css-extract-plugin/commit/e836a00cb164fecdb49e670e0552475477fa7994

PR #636, opened 2022-01-17
feat(plugin-external-link-icon): add locales option

Mister-Hope:


packages/@vuepress/theme-default/src/shared/options.ts
```diff
@@ -28,7 +29,7 @@ export interface DefaultThemePluginsOptions {
   /**
    * Enable @vuepress/plugin-external-link-icon or not
    */
-  externalLinkIcon?: boolean
+  externalLinkIcon?: ExternalLinkIconPluginOptions | boolean
```
meteorlxy, line 32: I'd prefer to keep `themePlugins` consistent, i.e. keep 	`externalLinkIcon?: boolean` here.

We can still keep the `openInNewWindow` locale config, and use a `resolveExternalLinkIconPluginOptions` util to resolve the options.

packages/@vuepress/plugin-external-link-icon/src/node/externalLinkIconPlugin.ts
```diff
@@ -2,11 +2,22 @@ import type { Plugin } from '@vuepress/core'
 import type { MarkdownEnv } from '@vuepress/markdown'
 import { path } from '@vuepress/utils'
 
-export type ExternalLinkIconPluginOptions = Record<never, never>
+export type ExternalLinkIconPluginOptions = {
+  /**
+   * i18n config
+   */
+  i18n: Record<string, string>
```
meteorlxy, line 9: Keep consistent with other plugins like `plugin-pwa-popup` or `plugin-search`.

```ts
locales?: LocaleConfig<{
  openInNewWindow: string
}>
```

meteorlxy (changes requested):

PR #638, opened 2022-01-18
feat: refine page types to control frontmatter type

Mister-Hope:


Mister-Hope:
I would like to have full control of `frontmatter` type when using `Page` and `PageData`.

This PR allows a new generic controlling type of frontmatter. Also there is a tweak in `Page`, and it should be better than the previous one.

packages/@vuepress/core/src/types/page.ts
```diff
@@ -6,141 +6,141 @@ import type { PageData, PageFrontmatter } from '@vuepress/shared'
...
-    htmlFilePath: string
-
-    /**
-     * Rendered html file path relative to dest directory
-     */
-    htmlFilePathRelative: string
-  }
+  ExtraPageFields extends Record<any, any> = Record<never, never>,
+  Frontmatter extends PageFrontmatter = PageFrontmatter
+> = PageData<ExtraPageFields, Frontmatter> & {
```
meteorlxy, line 11: `ExtraPageFields` may not be part of `PageData`.

`page.foo` and `page.data.foo` are different

meteorlxy (changes requested):

packages/@vuepress/core/src/types/page.ts
Mister-Hope: I understand your idea, but I am not seeing any difference between this change. The "Page Data is actually `page.data` on behind lines.

If you really want to declare a difference here, maybe you should copy all the fields in `PageData` and place here the second time. If you just extend it, then I would just regard it as an interface with no special meanings (Just a shape of some basic page properties), so that's why I place `ExtraPageFields` on it.

Mister-Hope: Or in another word, we rename current `PageData` as `BasicPage`

And we decalre `type PageData<...> = BasicPage<...>`

Then we wrote

```
type Page<...> = BasicPage & ExtraPageFields & {
  ...
  data: PageData<...>
}
```

Mister-Hope: I update the types to try making a better declaration.

PR #655, opened 2022-01-25
Sidebar Example - Locale

bacon-delight:
Added an example for rendering different sidebar options based on locale


meteorlxy:
Thanks, but the recommended way is to put it under different locales.

```
themeConfig: {
  locales: {
    '/': {
      sidebar: ...
    },
    '/zh/': {
      sidebar: ...
    },
  },
}
```

PR #657, opened 2022-01-26
feat(theme-default): fix title style when no content in custom container (close #648)

### LINKED ISSUE #648: [Feature request] Centered up and down title when no content in custom container
### Description

When there's no content in custom container, the title of custom container is not centered up and down. I don't think it's beautiful
image
If title can centered up and down, it will look better.
image

zh-lx:
fixed #648 
When there's no content in custom container, the title of custom container is not centered up and down, and the reason is that `margin-bottom: -0.4em;` overwrites the original `margin-bottom` of the `p` tag. So removing this style ensures that the title is center when no content in custom container. In addition, the container is not affected when it has content.

style of container without content:
image

style of container with content:
image

### DIFF
--- a/packages/@vuepress/theme-default/src/client/styles/custom-container.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/custom-container.scss
@@ -4,6 +4,9 @@
 
   .custom-container-title {
     font-weight: 600;
+    &:not(:only-child) {
+      margin-bottom: -0.4rem;
+    }
   }
 
   &.tip,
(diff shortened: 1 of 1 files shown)


meteorlxy:
The style will be slightly different without the offset. You'd better keep it consistent.

packages/@vuepress/theme-default/src/client/styles/custom-container.scss
```diff
@@ -4,7 +4,6 @@
 
   .custom-container-title {
     font-weight: 600;
-    margin-bottom: -0.4rem;
   }
```
Mister-Hope, line 7:
```suggestion
    &:not(:only-child) {
      margin-bottom: -0.4rem;
    }
  }
```

Mister-Hope (changes requested):

Mister-Hope (approved):

meteorlxy (approved):

PR #690, opened 2022-02-13
docs: fix wrong npm package link #690)

ostwindli:


meteorlxy:
Thanks for catching that ❤️

PR #698, opened 2022-02-17
feat(theme-default): subtitle automatically appears in the view when it overflows the visual area

### LINKED ISSUE #658: [Feature request]: subtitle automatically appears in the view
### Description

When I scroll markdown content, I need to swipe sidebar to see the corresponding subtitle if the current subtitle is too far behind：

image

### Proposed Solution

I want the subtitle to automatically appear in the view：
image

I have implemented this feature in my vuepress theme and can submit PR for this feature if needed.

zh-lx:
close #658 
When I scroll markdown content, I need to swipe sidebar to see the corresponding subtitle if the current subtitle is too far behind：
image

This PR will realize the function that the page content scrolls, if the active subtitle overflows the visual area, it will automatically scroll to the visual area. The effect is shown in GIF below:
auto-scroll


Mister-Hope (approved):

Mister-Hope:
Wait... Does it work with codegroup or even nested codegroup?

zh-lx:
> Wait... Does it work with sidebar group or even nested sidebar group? Since each sidebar group will have a sideLinks as children, there might be mutiple components in a single sideabr

Yes, I have tested the case, there might are many active sidebars but only one of them whose href is completely match both path and hash of the page. And the code will find this one and let it scrollIntoView.

packages/@vuepress/theme-default/src/client/components/SidebarItems.vue
```diff
@@ -1,8 +1,47 @@
 <script setup lang="ts">
 import SidebarItem from '@theme/SidebarItem.vue'
+import { watch } from 'vue'
+import { useRouter } from 'vue-router'
 import { useSidebarItems } from '../composables'
 
 const sidebarItems = useSidebarItems()
+
+const route = useRouter()
```
meteorlxy, line 9: The variable name `foo` should be consistent with `useFoo`. i.e. `const router = useRouter`

packages/@vuepress/theme-default/src/client/components/SidebarItems.vue
```diff
@@ -1,8 +1,47 @@
...
+
+watch(
+  () => route,
+  (router) => {
```
meteorlxy, line 13: Maybe we should only watch the hash?

packages/@vuepress/theme-default/src/client/components/SidebarItems.vue
```diff
@@ -1,8 +1,47 @@
...
+    // the sidebarItem dom whose href equals to the current href of the page
+    const activeSidebarItem = document.querySelector(
+      `[href="${router.currentRoute.value.path}${router.currentRoute.value.hash}"]`
```
meteorlxy, line 16: Could it be better to be more specific? For example, use `.sidebar sidebar-item a[href=...]`? There might be some other links like TOC

meteorlxy (changes requested):

packages/@vuepress/theme-default/src/client/components/SidebarItems.vue
zh-lx: I have changed it.

packages/@vuepress/theme-default/src/client/components/SidebarItems.vue
zh-lx: I have optimized this code.

packages/@vuepress/theme-default/src/client/components/SidebarItems.vue
zh-lx: I have optimized this code.

packages/@vuepress/theme-default/src/client/components/SidebarItems.vue
```diff
@@ -1,8 +1,44 @@
...
+import { watch } from 'vue'
+import { useRouter } from 'vue-router'
 import { useSidebarItems } from '../composables'
 
 const sidebarItems = useSidebarItems()
+
+const router = useRouter()
+
+watch(
+  () => router.currentRoute.value.hash,
```
Mister-Hope, line 12: I am not sure, could you explain why are you using `router` instead of `route`?

Why don't you choose:

```
const route = useRoute();

watch(() => route.hash, ...
```

zh-lx: Yes, you're right! I should use `useRoute` to replace `useRouter` and I have changed it.

Mister-Hope (approved):

meteorlxy (approved):

PR #720, opened 2022-02-26
fix(theme-default): only show toggleSidebarButton when necessary, fix…

Mister-Hope:
… #696


meteorlxy:
The toggle sidebar button is also used for navbar 🤔

Mister-Hope:
🤣 Thanks reminding that because I moved navbar items to a new place in my theme.
image

Mister-Hope:
Then it will need a refactor here, maybe I should make navbar items global provided like sidebaritems.

Mister-Hope:
Closing it first.

I will try to make some refactors to improve this.

Just realized that gestures can invoke sidebar either.

Also, here is a werid logic:

- When users disable sidebar, they won't find the toggle sidebar button to show sidebar on mobile with a valid sidebar config.

PR #731, opened 2022-03-01
fix(markdown): add aria-hidden for line-numbers

Mister-Hope:


packages/@vuepress/markdown/src/plugins/codePlugin/codePlugin.ts
```diff
@@ -109,10 +109,10 @@ export const codePlugin: PluginWithOptions<CodePluginOptions> = (
     if (useLineNumbers) {
       // generate line numbers code
       const lineNumbersCode = lines
-        .map((_, index) => `<span class="line-number">${index + 1}</span><br>`)
+        .map((_, index) => `<span class="line-number" >${index + 1}</span><br>`)
```
meteorlxy, line 112: remove extra space

meteorlxy (changes requested):

packages/@vuepress/markdown/src/plugins/codePlugin/codePlugin.ts
Mister-Hope: /doge Did not notice that

PR #736, opened 2022-03-01
feat(theme-default): enhance header dropdown experience

Shigma:
This pull request introduces different transition behaviors for header and aside navbars:

- header: transform & opacity, just like https://vuejs.org
- aside: collapse (DropdownTransition)

### DIFF
--- a/packages/@vuepress/theme-default/src/client/components/Navbar.vue
+++ b/packages/@vuepress/theme-default/src/client/components/Navbar.vue
@@ -67,7 +67,7 @@ function getCssValue(el: HTMLElement | null, property: string): number {
 
     <div class="navbar-items-wrapper" :style="linksWrapperStyle">
       <slot name="before" />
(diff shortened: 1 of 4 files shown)


Mister-Hope:
A serious problem, did you fully test it? Desktop mode is not working. Please make sure that `mouseleave` event is correct (I don't think that's correct) @Shigma

Mister-Hope:
https://user-images.githubusercontent.com/33315834/160976786-ede136c6-872f-418d-bc7c-f42bbed160be.mp4

I am afraid that if the dropdown menu on the right is too long, it will prevent the left dropdown to work

Mister-Hope:
I will revert this as this is a serious issue, the dropdown menu are clickable(but not visible)

image


You may check my theme: http://vuepress-theme-hope.github.io/v2/ to find a good way achieving this.

PR #764, opened 2022-03-20
Update example styles

AustinGil:
I think it's important to share examples that promote best practices. This change accounts for user preference as well as users using the find-in-page feature. 

See https://css-tricks.com/fixing-smooth-scrolling-with-find-on-page/

### DIFF
--- a/docs/reference/default-theme/styles.md
+++ b/docs/reference/default-theme/styles.md
@@ -23,8 +23,10 @@ The path of style file is `.vuepress/styles/index.scss`.
 You can add extra styles here, or override default styles:
 
 ```scss
-:root {
-  scroll-behavior: smooth;
+@media (prefers-reduced-motion: no-preference) {
+  :root:focus-within {
+    scroll-behavior: smooth;
+  }
 }
 ```
 


Mister-Hope:
https://user-images.githubusercontent.com/33315834/159173332-d474a1a3-0f03-4307-af28-c1c4096d059a.mp4

image
image

Mister-Hope:
> I think it's important to share examples that promote best practices. 

I dount what's a "best pratice", you cannot say what's better between having nicer behavior and greater compatablity

Mister-Hope:
I think that it's good for us to give a better example here, but I believe the original ones could be kept as well, also you may need some comment to have a little explain what you are doing. I don't think users are happy to see a much more complicated demo.

AustinGil:
On one hand, smooth scrolling is an enhancement. If we break it for some users that dont match compatibility, the fallback is default browser behavior. On the other hand, if we force smooth scrolling on users that have vestibular motion disorders, we may introduce accessibility concerns. So in this case, "best practice" was a convenient short way of saying that I think it's better to avoid introducing accessibility issues for some users at the cost of other users missing out on smooth scrolling.

But I dont want to miss the main point here. Docs on a project like this have the potential to reach a lot of folks, and I think it's good to be responsible with examples. To that point, I think my solution is responsible, but you're right that it does too much and raises questions. So maybe the best solution is to avoid smooth scrolling all together. After all, the documentation here is about style files. It can literally be just about anything.

Mister-Hope:
Hi, could you please upgrade this file?

https://github.com/vuepress/vuepress-next/blob/main/docs/.vuepress/styles/index.scss

I may edit the docs exmaple to a more simple one.

AustinGil:
Did you want that edit to go in this PR or in a new PR? 

Should I make a change to the CSS I had in the docs?

Mister-Hope:
Just editing docs style should be fine. I will use a another simple css snippet as  demo instead.

meteorlxy:
Thanks anyway

PR #771, opened 2022-03-22
feat(plugin-docsearch): add debounce in docsearch, close #216

### LINKED ISSUE #216: [Feature request] Debounce in docsearch
## Feature request

Currently, every type change in input box will send a search request, so the result list earlier will finally display if it's ending at last.
Also this cause the result list shrink when typing. I think a debounce with delay should be added.

Mister-Hope:


packages/@vuepress/plugin-docsearch/package.json
```diff
@@ -38,6 +38,7 @@
     "@vuepress/core": "2.0.0-beta.36",
     "@vuepress/shared": "2.0.0-beta.35",
     "@vuepress/utils": "2.0.0-beta.35",
+    "lodash.debounce": "^4.0.8",
```
meteorlxy, line 41: use `ts-debounce` as other packages

meteorlxy (changes requested):

PR #785, opened 2022-03-31
docs: format docs

Mister-Hope:


Mister-Hope:
Some of the code blocks are reformatted because the lines are too long on mobile devices, and some of the code blocks aren't (becuase they do look better before formatting).

Also I notice an issue here, it seems that you prefer not  indent code blocks in list (maybe you want to make code blocks look better on mobile), but that do have new issues. The code block will break order list number, and the following text won't be in the same `<li>` as expected, so they won't indent.

Here is an example:

image

The two `Then the perma...` should be inside li, and they are expected to be indented with 2 spaces, but they cannot because of the code block. (No matter whether you indent 2 space for them or not)

If you indent with 4 spaces, they will be rendered as code block:

image


So to make it right, we actually should indent the code block as expected:

image


I am not bargin something here, but I think making the struture right should be the first thing. If we do want the code block look better on mobile, we can add some styles. e.g:

```css
li > div[class=language-*] {
  /* remove indent here */
}
```

Instead of doing some hacky things in markdown which spoil the html structure.

meteorlxy:
Closing as too many conflicts

PR #807, opened 2022-04-09
feat(core): support package path in client files

Mister-Hope:


Mister-Hope:
If you think this change is ok, I will update test files for it.

meteorlxy:
Should avoid that as we discussed

Issue #816, opened 2022-04-14
Add css variables for font size


hyp530:
### Description

Wondering if you could provide css variables in `vars.scss' and `vars-dark.scss` for font-size -  both headers and body text.

Mister-Hope:
I don't think this is a common case, as most website font size is all 16px.

If you really want to modify this:

`.vuepress/styles/index.scss`

```scss
body {
  font-size: 18px !important;
}
```

PR #817, opened 2022-04-14
feat(theme-default): nomalize rem size

Mister-Hope:


Mister-Hope:
@meteorlxy  Inspired by #816, following root font size have better a11y.

packages/@vuepress/theme-default/src/client/styles/normalize.scss
```diff
@@ -14,7 +14,7 @@ body {
   font-family: var(--font-family);
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
-  font-size: 16px;
+  font-size: 1rem;
```
meteorlxy, line 17: The `normalize.scss` file is to ensure most elements to have normalized styles, instead of browser-specific styles. If changing this to `1rem`, it will depend on the browser's preset.

Mister-Hope:
Then how about adding an addtion style:

```css
: root {
  font-size: 16px;
}
```

packages/@vuepress/theme-default/src/client/styles/normalize.scss
Mister-Hope: But actually setting font-size to body has no effect on `rem`, if user set root font size with `32px`, VuePress still render a SUPER big heading 1.

meteorlxy:
What about simply moving `font-size: 16px;` to the `html, body` section? What's the point to specify `1rem` for body?

Mister-Hope:
Setting both as 16 px, user will fail to change body font-size when simply setting a new rem size.

PR #818, opened 2022-04-14
fix(markdown): resolve srcset attr for image (fix #809)

### LINKED ISSUE #809: [Bug report] img's srcset not handled correctly
## Description

Correct HTML code using `srcset` attribute for adaptive images in markdown file:

```markdown
<img src="../assets/blocks/foo.png" srcset="../assets/blocks/foo2.png 2x">
```

Gives an error:

```
[plugin:vite:import-analysis] Failed to resolve import "../assets/blocks/foo2.png" from ".vuepress/.temp/pages/en/blocks.html.vue". Does the file exist?
[/XXX/docs/.vuepress/.temp/pages/en/blocks.html.vue:7:25]()

...

6  |  import _imports_3 from '@source/assets/blocks/foo.png'
7  |  import _imports_4 from '../assets/blocks/foo2.png'
   |                          ^

...
```

Both files actually exist.

I guess the problem is that VuePress prepends the `src` value with `@source/` prefix, but doesn't do it to the `srcset` values.

## Versions

```
vuepress/core@2.0.0-beta.38 vuepress/cli@2.0.0-beta.38 darwin-arm64 node-v17.5.0
```

Mister-Hope:


.vscode/settings.json
```diff
@@ -9,6 +9,10 @@
     "files.trimTrailingWhitespace": false
   },
   "typescript.tsdk": "node_modules/typescript/lib",
+  "jest.autoRun": {
+    "enabled": false,
+    "watch": false
+  },
```
meteorlxy, line 15: Don't commit this 🤔

packages/@vuepress/markdown/__tests__/plugins/assetsPlugin.spec.ts
```diff
@@ -189,6 +190,41 @@ describe('@vuepress/markdown > plugins > assetsPlugin', () => {
       '<img src=".../汉字.png">',
       '<img src=".../100%.png">',
       '<img alt="attrs" src=".../attrs.png" width="100px">',
+
+      /* srcset */
+
```
meteorlxy, line 195: extra empty line

packages/@vuepress/markdown/__tests__/plugins/assetsPlugin.spec.ts
```diff
@@ -189,6 +190,41 @@ describe('@vuepress/markdown > plugins > assetsPlugin', () => {
...
+      // relative paths
+      '<img srcset="./foo.png 1x, ../sub/foo.png 2x, ./foo/bar.png 1024w ,../sub/foo/bar.png 2048w, ../baz.png 4096w,../../out.png">',
+      '<img srcset="./汉字.png 1x, ./100%.png">',
+      '<img alt="attrs" srcset="./attrs.png" width="100px">',
+      // aliases
+      '<img srcset="@alias/foo.png 1x,@alias/汉字.png 2x, @alias/100%.png 3x">',
```
meteorlxy, line 201: Missing whitespace after the first comma.

It could be better to add extra cases without the comma, e.g.:

```ts
// no whitespace after comma
'<img .....'
```

packages/@vuepress/markdown/__tests__/plugins/assetsPlugin.spec.ts
```diff
@@ -189,6 +190,41 @@ describe('@vuepress/markdown > plugins > assetsPlugin', () => {
...
+      // invalid paths
+      '<img srcset=".../invalid.png 1x, .../汉字.png 2x, .../100%.png 3x">',
+      '<img alt="attrs" srcset=".../attrs.png 1x, .../default.png" width="100px">',
+      // invalid srcset
+      '<img srcset="../invalid.png, ../汉字.png, .../100%.png 3x">',
+
+      /** both */
+      // relative paths
+      '<img srcset="./foo.png 1x, ../sub/foo.png 2x, ./foo/bar.png 1024w ,../sub/foo/bar.png 2048w, ../baz.png 4096w,../../out.png 3x" src="./default.png">',
+      '<img src="./100%.png"  srcset="./汉字.png 1x" >',
```
meteorlxy, line 219: Extra whitespace before `srcset`. Check the format again

meteorlxy (changes requested):

packages/@vuepress/markdown/__tests__/plugins/assetsPlugin.spec.ts
Mister-Hope: That's expected, because that unit test is to mixed space position and see if it works. I am not sure whether it's necessary to add a lot of combination into the unit test.

packages/@vuepress/markdown/__tests__/plugins/assetsPlugin.spec.ts
Mister-Hope: That's expected in my design either, users may not  have standard input.

meteorlxy: Then you need to add other cases for that purpose.

meteorlxy: For example, `non-formatted` to mix those cases.

Mister-Hope: Emm, well, I will add those. Just think that it was not really necessary to add a lot of test cases which are acually unneeded.

.vscode/settings.json
Mister-Hope: The poblem is that vscode is not providing a `.vscode/private.setting.json` thing. And you are commiting your own settings to the repo as well.

Mister-Hope: If you think that adding this file to `.gitignore` and removin it is ok. I will make a commit to do that.

Mister-Hope:
@meteorlxy  Updated. Add another test case because markdown-it will "pre mark" any mutiple line element as "block", so they can not be rendered as inline item in the original test.

meteorlxy:
Good job :+1:. Syntax & formatting is important in markdown tests.

packages/@vuepress/markdown/__tests__/plugins/assetsPlugin.spec.ts
```diff
@@ -235,6 +273,40 @@ describe('@vuepress/markdown > plugins > assetsPlugin', () => {
...
+          '<img srcset=".../invalid.png 1x, .../汉字.png 2x, .../100%.png 3x">',
+          '<img alt="attrs" srcset=".../attrs.png 1x, .../default.png" width="100px">',
+          // invalid srcset
+          '<img srcset="@source/invalid.png, @source/汉字.png, .../100%.png 3x">',
+
+          /* both */
+
+          // relative paths
+          '<img srcset="@source/sub/foo.png 1x, @source/sub/foo.png 2x, @source/sub/foo/bar.png 1024w, @source/sub/foo/bar.png 2048w, @source/baz.png 4096w, @source/../out.png 3x" src="@source/sub/default.png">',
+          '<img src="@source/sub/100%.png" srcset="@source/sub/汉字.png 1x" >',
```
meteorlxy, line 303: extra whitespace

packages/@vuepress/markdown/__tests__/plugins/assetsPlugin.spec.ts
```diff
@@ -396,5 +540,126 @@ describe('@vuepress/markdown > plugins > assetsPlugin', () => {
         )
       })
     )
+
+    const source2 = [
```
meteorlxy, line 544: Why not put them inside `source` and `testCases` directly?

Mister-Hope: already comment the reason

meteorlxy (approved):

PR #819, opened 2022-04-14
feat(markdown): update line-number

Mister-Hope:


Mister-Hope:
Still, another improvement for "Read mode"

Mister-Hope:
BTW, this css style is quite safe to use:

image

meteorlxy:
> The previous one still not work well in some browsers.

Any repro/screenshots? Modifying the markdown plugin will affect all community themes who rely on it.

Mister-Hope:
yep，but I think rare plugin would do something to line number or highlight
 part.l, since I am not editing the outer structure, so they should still be able to select the part they want.

At least my plugin would work fine.

I will provide a screenshot later as you request.

Mister-Hope:
Before:

image

After:

image

Mister-Hope:
Like I said, there is acually no difference.

meteorlxy:
> Like I said, there is acually no difference.

I was asking for the differences / improvements 😅

PR #820, opened 2022-04-14
fix: fix: file does not exist
fixed these:

 error client file does not exist:
*/vuepress-theme-gungnir/packages/theme/src/client/clientAppEnhance.js

izhaong:


Mister-Hope:
Default theme is compiled to js and output to lib when pubshing, make sure to use the correct file

PR #821, opened 2022-04-14
fix: file does not exist

izhaong:
fixed these:
error client file does not exist: */vuepress-theme-gungnir/packages/theme/src/client/clientAppEnhance.js


Mister-Hope:
Default theme is compiled to js and output to lib when pubshing, make sure to use the correct file

PR #822, opened 2022-04-14
fix: file does not exist

izhaong:
fixed these:
error client file does not exist: */vuepress-theme-gungnir/packages/theme/src/client/clientAppEnhance.js

### DIFF
--- a/packages/@vuepress/theme-default/src/node/defaultTheme.ts
+++ b/packages/@vuepress/theme-default/src/node/defaultTheme.ts
@@ -67,10 +67,10 @@ export const defaultTheme: Theme<DefaultThemeOptions> = (
 
     clientAppEnhanceFiles: path.resolve(
       __dirname,
-      '../client/clientAppEnhance.js'
+      '../client/clientAppEnhance.ts'
     ),
 
-    clientAppSetupFiles: path.resolve(__dirname, '../client/clientAppSetup.js'),
+    clientAppSetupFiles: path.resolve(__dirname, '../client/clientAppSetup.ts'),
 
     extendsPage: (page: Page<DefaultThemePageData>) => {
       // save relative file path into page data to generate edit link


Mister-Hope:
Default theme is compiled to js and output to lib when pubshing, **make sure to use the correct file**

Mister-Hope:
# Final Warning, I will block you if you continue posting dumplicated and unneeded PRs.

izhaong:
> Default theme is compiled to js and output to lib when publishing, **make sure to use the correct file**

Sorry, I repeatedly submit it. But the file path I modified is right.

Mister-Hope:
> Default theme is compiled to js and output to lib when publishing, make sure to use the correct file

Is there any trouble? 

Translation: 默认主题会编译为js并在发布时输出到lib文件夹，请确保使用正确的文件

Mister-Hope:
Or:

### Do use `@vuepress/theme-default/lib/xxx`

### Do NOT use  `@vuepress/theme-default/src/xxx`

izhaong:
好吧  我只是在本地环境开发主题时对这个路径报错感到困惑 所以提了pr  没有恶意

Mister-Hope:
If the extension changed to `.ts` then the theme won't work becuase there an't any ts files in `lib` folder.

Some suggestions:

- if you only want to slighty modify default theme, extend it and use lib folder should be ok
- if you want to create your new theme, you may want to start with default theme, you can just copy the source files to your directory and start your jorney (also remember following MIT licnese)

Also, just a reminder, most features for documatation already have related plugins and you should be able to find them at awesome-vuepress repo. Also, `vuepress-theme-hope` is powerfull enough for you in most cases (unless you want a blog theme with different outlook) and it will be released as `@vuepress/theme-pro` in future.

izhaong:
I got it. Thanks for your patience.

PR #827, opened 2022-04-16
fix(plugin-search): fix hotkey when clicking any element on page (fix #746)

### LINKED ISSUE #746: [Bug report] 点击侧边栏，搜索快捷键失效
展示站
<https://algorithm-pattern.vercel.app/>

源代码
<https://github.com/lyj0309/algorithm-pattern>

如题，点击侧边栏后，使用‘/’快捷键搜索，无法跳到搜索栏

Mister-Hope:


Mister-Hope:
@meteorlxy I am not sure why the original one force the target to be body, but clicking any anchor element or even clicking a haeding or paragrah, will cause the next keypress has a target of that element, so the hotkey actually only works after refreshing page with no action (except scrolling) now. And I think that's not the right behavior.

If you are worrying about people may use the `s` key for something, we may add ctrl key to the default hot key.

meteorlxy:
That's migrate directly from v1

packages/@vuepress/plugin-search/src/client/composables/useHotKeys.ts
```diff
@@ -10,7 +10,7 @@ export const useHotKeys = ({
 }): void => {
   const onKeydown = (event: KeyboardEvent): void => {
     if (!input.value || hotKeys.value.length === 0) return
-    if (event.target === document.body && hotKeys.value.includes(event.key)) {
+    if (hotKeys.value.includes(event.key)) {
```
meteorlxy, line 13: Will this be triggered when user typing `s` in the search input?

Mister-Hope: Maybe, even not, I will suggest adding `Ctrl` key. I will update this PR. Opening it is just for having a place to ask the above question.

Mister-Hope:
Rebuild with #831

PR #837, opened 2022-04-19
feat(core): ensure important app options is correct

Mister-Hope:


Mister-Hope:
Some noob users may set `base: 'xxx'` and `dest : './'` and this is REALLY common, so it's good for us to avoid that. Because the first one breaks the whole app ,and the second one will empty user source file (and even `.git`)

Mister-Hope:
Any suggetions with this PR?

PR #848, opened 2022-04-24
chore: compatable with pnpm@v7

Mister-Hope:


meteorlxy:
We'll update to pnpm v7 after it's officially released. Now use v6 instead.

Mister-Hope:
I think it's fine to convert this to draft as we will need this.

pnpm will release v7 very soon after node 12 moves out of LTS

PR #852, opened 2022-04-27
build: drop support for node 12

Mister-Hope:
BREAKING CHANGES: node 12 is dropped

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -2,7 +2,7 @@
 
 ## Prerequisites
 
-- [Node.js v12+](https://nodejs.org/)
+- [Node.js v14+](https://nodejs.org/)
 - [Yarn v1 classic](https://classic.yarnpkg.com/en/) (Optional)
 
 ::: tip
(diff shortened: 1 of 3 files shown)


Mister-Hope:
Node 12 will end up LTS support in 2022/5/1, and target: es2020 is aligning with minimal node 14 support.

This helps generating code with less polyfills.

meteorlxy (changes requested):
As commented above

docs/guide/deployment.md
```diff
@@ -52,7 +52,7 @@ jobs:
         uses: actions/setup-node@v1
         with:
           # choose node.js version to use
-          node-version: '14'
+          node-version: '16'
```
meteorlxy, line 55: No need to change the deployment guide. Use the lowest LTS version is OK here. (We should use 12 before :sweat_smile:)

meteorlxy (changes requested):

meteorlxy (approved):

PR #853, opened 2022-04-27
chore: remove settings.json

Mister-Hope:


Mister-Hope:
Removing settings.json is needed, as contributors like me would have personal workspace settings to work well, but changes are not accepted, also current setting.json is mixing personal taste extension settings like cSpell which is not necessary for the project.

meteorlxy:
There are both pros and cons to include or exclude settings inside a repo. Excluding it is also a personal taste

Issue #857, opened 2022-04-29
[Feature request] add content-bottom slot to support comment components 


ManfredHu:
### Clear and concise description of the problem

The request to add a comment component at the bottom of the article is very difficult to deal with in the existing way.

According to the [document](https://v2.vuepress.vuejs.org/reference/default-theme/extending.html#layout-slots), the bottom content will be inserted here in 1 instead of 2.

<img width="872" alt="image" src="https://user-images.githubusercontent.com/9734888/165896795-904d4241-89fa-4a52-b86d-db8e94363f1d.png">
<img width="747" alt="image" src="https://user-images.githubusercontent.com/9734888/165896864-b6597e00-171d-4abf-a73f-dfef526ac041.png">

You need to add an extra slot for the real area at the bottom of the article

### Suggested solution

<img width="693" alt="image" src="https://user-images.githubusercontent.com/9734888/165897172-a8372cea-e632-415e-a020-70eeb5b1bbb0.png">

add a slot name "content-bottom"

Mister-Hope:
Reasonable, pr welcome (I suppose you can do that)

PR #858, opened 2022-04-29
feat(theme-default): add content-bottom slot

ManfredHu:
Associate with #857


Mister-Hope (changes requested):
Also you may add a content-top as well, e.g. for breadcrumbs

PR #859, opened 2022-04-29
feat: add vite-merge package

Mister-Hope:


meteorlxy:
You could consider publishing it in another repo.

PR #861, opened 2022-04-30
feat(core): support frontmatter.redirectFrom

shigma:
In many scenarios, developers have modified the structure of the document, resulting in old links inaccessible. We need a way to specify a set of legacy paths for each file. When the user visits the corresponding path, he will be automatically redirected to the new path.

### DIFF
--- a/docs/reference/frontmatter.md
+++ b/docs/reference/frontmatter.md
@@ -184,6 +184,14 @@ permalinkPattern: :year/:month/:day/:slug.html
   - [Frontmatter > permalink](#permalink)
   - [Node API > Page Properties > permalink](./node-api.md#permalink)
 
+## redirectFrom
+
+- Type: `string[]`
+
+- Details:
+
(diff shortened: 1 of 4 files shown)


Mister-Hope:
Instead of this pr, I strongly recommend you to use `vuepress-plugin-redirect2`.

It has much more better seo

Shigma:
@Mister-Hope 

I investigated your plugin and it's really well done. The scheme of writing redirect in markdown also does provide better SEO support. But it is a bit troublesome to maintain a batch of empty files for the purpose of refactoring. It would be better if it could support `redirectFrom` configuration. Anyway I'm going to migrate to your plugin when you support vuepress from beta.40+.

BTW, if you want to have `vuepress-plugin-redirect` instead of "2", I'm willing to give you a maintainer. anyway, I don't maintain it anymore.

Mister-Hope:
> redirectFrom and redirectTo

I will support that, that’s a nice idea.

I am currently postgraduate student and study theoretical physic, so busy this week. May have that support at weekend.

> BTW, if you want to have vuepress-plugin-redirect instead of "2", I'm willing to give you a maintainer. anyway, I don't maintain it anymore.

That’s great

Mister-Hope:
> redirectFrom and redirectTo

These are supported, so I think we can close this PR.

PR #876, opened 2022-05-06
fix: @vuepress/client dependency in vuepress-vite

nknapp:
The dependency "vuepress-vite -> @vuepress/client" should not be hardcoded (as far as I understand), but always use the in-workspace version (i.e. "workspace:*")

### DIFF
--- a/packages/vuepress-vite/package.json
+++ b/packages/vuepress-vite/package.json
@@ -39,7 +39,7 @@
     "@vuepress/theme-default": "workspace:*"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


meteorlxy:
It's a known issue of pnpm https://github.com/pnpm/pnpm/issues/4164

So I have to rollback the workspace protocol in 8230e24e03b611cc9a7b3aa8fa64483d2af6a3a3

nknapp:
oh, thanks for the info. I just came across this, because I had peer-dependency warnings in my project. But I wasn't able to create a minimal example for that.

PR #877, opened 2022-05-07
fix: show the search modal correctly

zpfz:
the search modal shows incorrectly in mobile when the page was scrolled down, so i found the problem what it cause.
```css
@media (max-width: 750px) {
  .DocSearch-Container{
    position: absolute;
  }
}
```
👇👇👇
```css
@media (max-width: 750px) {
  .DocSearch-Container{
    position: fixed;
  }
}
```
keeping the default css can solve this problem.

### DIFF
--- a/packages/@vuepress/plugin-docsearch/src/client/components/Docsearch.ts
+++ b/packages/@vuepress/plugin-docsearch/src/client/components/Docsearch.ts
@@ -9,6 +9,8 @@ import { useDocsearchShim } from '../composables'
 
 import '@docsearch/css'
 
+import '../styles/docsearch.css'
+
 export const Docsearch = defineComponent({
   name: 'Docsearch',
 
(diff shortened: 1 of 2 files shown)


meteorlxy:
Oops, thanks for catching it, didn't be aware of that.

I wonder if this issue only occurs in our default theme?

- If it's a default-theme-only bug, we'd better fix the style of default theme instead.
- If it's a common bug, we'd better submit a PR to the docsearch repo directly.

Mister-Hope:
this issue happened in my theme and I already fixed it.

zpfz:
> Oops, thanks for catching it, didn't be aware of that.
> 
> I wonder if this issue only occurs in our default theme?
> 
> - If it's a default-theme-only bug, we'd better fix the style of default theme instead.
> - If it's a common bug, we'd better submit a PR to the docsearch repo directly.
> 
> 

All right...maybe i need to revoke this issue~it's DocSearch's problem.

Issue #729, opened 2022-02-28
[Feature request] Use better commitizen adapter <cz-git>


Zhengqbbb:
### Description

**I open sourced a commitizen adapter [cz-git](https://github.com/Zhengqbbb/cz-git). It is perfect for projects using `monorepo project`**

- Supports search and selection on the command line, reducing spelling errors.
- Highly Customizable, but the output format follows the standard [Angular commit](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits) specification.
- [Better for monorepo engineering](https://github.com/Zhengqbbb/cz-git#scopes) and commitlint project to give relevant verification information to the command line.

![demo](https://user-images.githubusercontent.com/40693636/156009970-3eee46a5-ed17-49cf-9d3d-e4622ee10d3f.gif)

### Proposed Solution

If the manager dares to be interested, I am very willing to provide PR.
I'm more than willing to make adjustments if there are suggestions.

Mister-Hope:
I am not sure about it, the tool is nice, but I am not seeing great benifit with it. You shall wait to see @meteorlxy 's opinion

The only thing the plugin provides is probably auto completing scope for this project (as we have long scopes). And that autocomplete based on input is nice.

> I'm more than willing to make adjustments if there are suggestions.

For the suggestions, IMO a such tool should be lightweight. I have a quick look at the lock file changes, and I found that your tool has more deps then I expected. So if there is any suggestions, (with your package rather than this PR), I would like to say maybe you can try to lower the deps number. And try require same versions of tools if possible.

Zhengqbbb:
@Mister-Hope Thank Hope ! 
> (Maybe comming from `inquirer-autocomplete-prompt` I guess?)

Yeah. I use it.

> I would like to say maybe you can try to lower the deps number. And try require same versions of tools if possible.

Great suggestion. That will be my next feature ~

PR #907, opened 2022-05-24
chore: use cz-git adapter

Zhengqbbb:
Hi I am the author of cz-git, 
After this issue #729 nearly two months of iterations, I think it's time for cz-git to be added to vuepress-next
In response to the comments last time, I have done a lot of rectification work.

- **Lightweight**:  only one dependency package cz-git. 1.7MB.
- **More custom feature**: More custom feature have been added on the basis of the original. Of course, I checked daily commit habits to add basic configure.
- **Improve development experience**. Below I will introduce

💡In addition to the feature that can be searched and selected up and down.
This time I think one ideal: **automatic pin top scope by obtaining scope in the git status folder path**
🤔So i have done it in this PR.

demo:
demo

Hope my opinion can be adopted, And enjoy it, of course, if you have a good idea, can tell me


meteorlxy:
The adapter is cool :+1:

However, we are not using commitizen in our repo (and not in other projects of Vue.js Team). So currently we'll not merge it.

PR #930, opened 2022-06-07
docs(register-components): correct example import

jshor:
The import has changed to export default.


meteorlxy:
Thanks, but named export is also allowed.

PR #980, opened 2022-07-08
Escape double-quotes in head tag attribute values

wm-noble:
To reproduce the bug:

```
// config.js
module.exports = {
  head: [
    ['script', {
       src: 'https://something.com/foo.js',
       'data-payload': '{"token": "hello"}',  // Use an attribute value with a JSON-encoded string.
    }],
  ],
```

You should run into the following error:

```
Document.querySelectorAll: 'head > script[src="https://something.com/foo.js"][data-payload="{"token": "hello"}"]' is not a valid selector
```

The problem is that the `data-payload` value is a JSON-encoded string, which requires double quotes, but the current implementation simply wraps the CSS selector for that call to `querySelectorAll` in double quotes and calls it a day. Use `JSON.stringify` instead. Since `value` is known to be a string in this context, this should always work (tested locally manually).

### DIFF
--- a/packages/@vuepress/client/src/setupUpdateHead.ts
+++ b/packages/@vuepress/client/src/setupUpdateHead.ts
@@ -81,7 +81,7 @@ export const queryHeadTag = ([
   const attrsSelector = Object.entries(attrs)
     .map(([key, value]) => {
       if (isString(value)) {
-        return `[${key}="${value}"]`
+        return `[${key}=${JSON.stringify(value)}]`
       }
       if (value === true) {
         return `[${key}]`


meteorlxy:
I think users should escape those chars by themselves in the `head` config 🤔

wm-noble:
Escaping is for string literals. It happens at compile time, not run time; there is no "second round" of escaping when you interpolate a string at run time, so this is impossible.

Mister-Hope:
We are always using double quotes everywhere, and double quotes is not allowed in a property value surrounded by double quotes. So to perform adding a valid tag, you are supported to escape them yourself. I think that's the central point meteorlxy express.

wm-noble:
The problem is that this is simply impossible when string interpolation is involved. Let me give you some examples to illustrate:

```
s = '{"a": "5"}';
`foo: "${s}"`                        // -> 'foo: "{"a": "5"}"'
s = '{\"a\": \"5\"}';
`foo: "${s}"`                        // -> 'foo: "{"a": "5"}"'
s = '{\\"a\\": \\"5\\"}';
`foo: "${s}"`                        // -> 'foo: "{\\"a\\": \\"5\\"}"'
s = '{\\\"a\\\": \\\"5\\\"}';
`foo: "${s}"`                        // -> 'foo: "{\\"a\\": \\"5\\"}"'
```

As you can see, no amount of backslashes will give you the result you want. You can verify this in your JS console. I only patched this one line since it solved the specific problem I was running into, but if you are indeed "using double quotes everywhere" then this problem will exist everywhere that a string containing double quotes is interpolated into a template literal that also contains double quotes. I could submit a larger PR if you'd like and try to get the rest of the cases, just want to make sure we're on the same page first.

wm-noble:
Lol I now realize where the confusion came from. Nevermind!

PR #985, opened 2022-07-15
add `alternate` head for i18n

Tinymemo:
## Background

According to Google's guidance, a website with multi languages or for multi region should use `hreflang`.

- <https://developers.google.com/search/docs/advanced/crawling/managing-multi-regional-sites>
- <https://developers.google.com/search/docs/advanced/crawling/localized-versions>

The method is to add `link`s to `<head>`:

```html
<link rel="alternate" hreflang="lang_code" href="url_of_page" />
```

For each language version, a `link` should be added. And the url must contains schema and hostname, `//example.com/foo` or `/foo` is not valid.

## Changes

- add a `url` to site data, for example `https://example.com`
- add `link`s to head when there exists more than one `locale`s
- the first `locale` will be the default language

## More discussion

- the name `url` comes from `hexo` config, but may be not so accurate
- the place of option: should it appear in `locale` options? (to support the situation: different language pages under different domain names)

### DIFF
--- a/packages/@vuepress/client/src/resolvers.ts
+++ b/packages/@vuepress/client/src/resolvers.ts
@@ -46,7 +46,8 @@ export const resolvers = reactive({
   resolvePageHead: (
     headTitle: PageHeadTitle,
     frontmatter: PageFrontmatter,
-    siteLocale: SiteLocaleData
+    siteLocale: SiteLocaleData,
+    alternate: PageHead
   ): PageHead => {
     const description = isString(frontmatter.description)
       ? frontmatter.description
@@ -56,6 +57,7 @@ export const resolvers = reactive({
       ...siteLocale.head,
       ['title', {}, headTitle],
       ['meta', { name: 'description', content: description }],
+      ...alternate,
     ]
     return dedupeHead(head)
   },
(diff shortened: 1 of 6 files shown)


Mister-Hope:
I think seo2 plugin support this already.

Could you check that?

Tinymemo:
@Mister-Hope Thanks for your tips. I missed the plugin. I'll try it.

Tinymemo:
@Mister-Hope I've tried, and found seo2 plugin didn't output `hreflang`, it just ouput some tags like this:

```html
<meta property="og:locale" content="zh-CN">
<meta property="og:locale:alternate" content="en-US">

<link rel="canonical" href="https://vuepress-theme-hope.github.io/v2/seo/zh/">
```

No `link[rel=alternate]` found.

So I will reopen this issue. Also, we can have more disscussions.

Mister-Hope:
You may open a new PR or issue there. `@vuepress/core` does not have any built-in SEO optimizes, and I think it's better to handle through a plugin.

Mister-Hope:
I supported this through https://github.com/vuepress-theme-hope/vuepress-theme-hope/commit/2e6b90ddacf625b77216918c33d5faaf03a905e5

Tinymemo:
@Mister-Hope Thanks for your efforts, seems very well, except for one thing, seems missing a `x-default`.

Anyway, I'll switch to your plugin `seo2`.

Issue #958, opened 2022-06-19
[Bug report] Title is too long


Nofated095:
### Description

When set a very long title for the project, the title can't hide correctly, and the button for the dark/light and search can been blocked

it sometimes happens on the mobile devices

![5C08DB81-6B9E-4F3D-A6C7-A80C6C062416](https://user-images.githubusercontent.com/49985975/174473075-c93b41be-7cb7-4898-9306-b3498534132c.png)

### Reproduction

https://ls-bot.9595095.xyz/

### Used Package Manager

yarn

### System Info

```shell
build correctly
```

Mister-Hope:
@meteorlxy  Here, alogillia search is breaking the style.

![image](https://user-images.githubusercontent.com/33315834/174543223-5e05312b-745f-49a5-b6f9-4bfcf25cf8aa.png)

I am willing to fix it, but I am not sure we should override this in docsearch plugin, or the default theme should fix this when it detects docsearch component?

meteorlxy:
Should be default theme scope

PR #987, opened 2022-07-16
fix(theme-default): samesite css

rottenpen:
# Description
Linked with this feature request issue #958 

# Note

mobile:
image
pc:
image

### DIFF
--- a/packages/@vuepress/theme-default/src/client/styles/navbar.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/navbar.scss
@@ -15,10 +15,15 @@
   }
 
   .site-name {
+    display: inline-block;
+    width: calc(100vw - 18rem);
     font-size: 1.3rem;
     font-weight: 600;
     color: var(--c-text);
     position: relative;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
   }
 
   .navbar-items-wrapper {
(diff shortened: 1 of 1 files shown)


meteorlxy:
Could you help to confirm if the fix valid when there is a search input?

For example, enabling `@vuepress/plugin-search` or `@vuepress/plugin-docsearch`

rottenpen:
its not good when there is a search input. and i fix it. @meteorlxy  
search input:
image
image
default:
image
image

meteorlxy:
I saw you set the width to a fixed value. Have you tested with navbar items?

rottenpen:
> I saw you set the width to a fixed value. Have you tested with navbar items?
what is navbar items ? multi navbar items?
or this ⬇️？
image

Mister-Hope:
image

I think your fix is invalid.

packages/@vuepress/theme-default/src/client/styles/navbar.scss
```diff
@@ -15,10 +15,15 @@
   }
 
   .site-name {
+    display: inline-block;
+    width: calc(100vw - 18rem);
```
Mister-Hope, line 19: Obviously because of dynamic contents in navbar, we should never set this to a fixed value.

Mister-Hope (changes requested):

packages/@vuepress/theme-default/src/client/styles/navbar.scss
rottenpen: oh，i see. Is it better to limit the navbar-items-wrapper and navbarBrand width with js?
conflict between the two when their too long

PR #1001, opened 2022-07-28
feat(plugin-google-analytics): add debug option

EdieLemoine:
Adds the ability to enable DebugView in development.

More information on DebugView: https://support.google.com/analytics/answer/7201382

### DIFF
--- a/docs/reference/plugin/google-analytics.md
+++ b/docs/reference/plugin/google-analytics.md
@@ -67,7 +67,7 @@ export default {
 - Example:
 
 ```js
-module.exports = {
+export default {
   plugins: [
     googleAnalyticsPlugin({
       id: 'G-XXXXXXXXXX',
--- a/docs/reference/plugin/google-analytics.md
+++ b/docs/reference/plugin/google-analytics.md
@@ -62,7 +62,7 @@ export default {
 
 - Details:
 
-  Set to true to enable sending events to DebugView in development. [More information on DebugView](https://support.google.com/analytics/answer/7201382)
+  Set to `true` to enable sending events to DebugView. [See more information on DebugView](https://support.google.com/analytics/answer/7201382).
 
 - Example:
 
--- a/docs/reference/plugin/google-analytics.md
+++ b/docs/reference/plugin/google-analytics.md
@@ -66,7 +66,7 @@ export default {
 
 - Example:
 
-```js
+```ts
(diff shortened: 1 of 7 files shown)


Mister-Hope (approved):

EdieLemoine:
What's left for this PR to be merged? I see the coverage has decreased, but this plugin wasn't covered at all before this change.

meteorlxy:
Sorry for delay. Didn't have time to look into this. I'll handle it after #1030 (a huge refactor :man_facepalming:)

packages/@vuepress/plugin-google-analytics/src/client/composables/useGoogleAnalytics.ts
```diff
@@ -41,5 +45,10 @@ export const useGoogleAnalytics = (id: string): void => {
   }
 
   gtag('js', new Date())
-  gtag('config', id)
+
+  if (options.debug) {
+    gtag('config', options.id, { debug_mode: true })
+  } else {
+    gtag('config', options.id)
+  }
```
meteorlxy, line 53: Can we use `gtag('config', options.id, { debug_mode: options.debug === true })` ? I'm not sure if setting `{ debug_mode: false }` will cause any different behavior

docs/reference/plugin/google-analytics.md
```diff
@@ -55,3 +55,24 @@ module.exports = {
...
 ```
+
+### debug
+
+- Type: `boolean`
+
+- Details:
+
+  Set to true to enable sending events to DebugView in development.
+
```
meteorlxy, line 66: It would be nice to add a `- Also see: ` and add the link to the debug mode docs.

meteorlxy (changes requested):
Could you help to resolve the conflicts with main branch?

packages/@vuepress/plugin-google-analytics/src/client/composables/useGoogleAnalytics.ts
EdieLemoine: That would make sense and look nice, but unfortunately they said this on the page I linked in the description:

> To disable debug mode, exclude the 'debug_mode' parameter; setting the parameter to 
 false doesn't disable debug mode.

meteorlxy (approved):

PR #1012, opened 2022-08-04
fix(plugin-docsearch): fix plugin-docsearch can not successful navigate

Fatpandac:
The fix can not use the key of `Enter` to select the search result to navigate successfully.
image


meteorlxy:
Oh, didn't notice this PR. Has already been fixed by b6ded16f607f5db5a27967260e7e6b7b69a26c2b

Thanks anyway ❤️

PR #1021, opened 2022-08-08
Add `r` to `$codeLang` variable

mihaiconstantin:
While using `VuePress` to build the documentation for my `R` package, I noticed that the language ID is missing from the top right corner (i.e., as seen below).

CleanShot 2022-08-08 at 16 20 49@2x

I can't recall seeing any mention in the official documentation on how to address this. The culprit seems to be `r` missing from the `$codeLang` `SCSS` variable:

https://github.com/vuepress/vuepress-next/blob/100fac96717fb5bf22c976f5c3ab7f454e821742/packages/%40vuepress/theme-default/src/client/styles/code.scss#L250-L256

After updating `$codeLang` to

```scss
$codeLang: 'c' 'cpp' 'cs' 'css' 'dart' 'docker' 'fs' 'go' 'html' 'java' 'js'
  'json' 'kt' 'less' 'makefile' 'md' 'php' 'py' 'rb' 'rs' 'sass' 'scss' 'sh'
  'styl' 'ts' 'toml' 'vue' 'yml' !default;
```

the language ID seems to be properly displayed, i.e.:

CleanShot 2022-08-08 at 16 27 14@2x

### DIFF
--- a/packages/@vuepress/theme-default/src/client/styles/_variables.scss
+++ b/packages/@vuepress/theme-default/src/client/styles/_variables.scss
@@ -7,5 +7,5 @@ $MQMobileNarrow: 419px !default;
 
 // code languages
 $codeLang: 'c' 'cpp' 'cs' 'css' 'dart' 'docker' 'fs' 'go' 'html' 'java' 'js'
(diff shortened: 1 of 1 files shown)


Mister-Hope:
This is a public project, and r is not a common language.

You should modify $codeLang variable yourself using `.vuepress/styles/palette.scss`

PR #1031, opened 2022-08-16
feat(core): add hook before initialize

EdieLemoine:
This adds a lifecycle hook before the app is initialized. I'm not sure if you want this implemented, 

I want to change/update locale data through a plugin, but I noticed the changes are not being reflected in the pages array in other plugins. I checked out the source and discovered this is because the earliest hook, `onInitialized` happens after pages are resolved. So I added another hook right before it.

I've updated the unit tests and tested it with my use case and it works like a charm.

### DIFF
--- a/docs/reference/plugin-api.md
+++ b/docs/reference/plugin-api.md
@@ -323,6 +323,14 @@ export default {
 
 ## Lifecycle Hooks
 
+### beforeInitialize
+
+- Type: `(app: App) => void | Promise<void>`
+
+- Details:
+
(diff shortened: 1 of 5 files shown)


meteorlxy:
I'm not sure if it is necessary because we can do things in pluing before initialization via "PluginFunction"

```js
const fooPlugin = (app) => {
  // here we can do things just like beforeInitialize

  return {
    name: 'vuepress-plugin-foo',
    onInitialized(app) {  /* ... */ }
  }
}
```

EdieLemoine:
@meteorlxy Wow, I blame tunnel vision. If having this as a hook doesn't have any other advantages this can be closed. :) Thanks

PR #1030, opened 2022-08-16
feat: migrate to pure ESM

meteorlxy:
- feat(cli): support .mjs config file
- feat(bundler-vite): support vite 3
- test: migrate from jest to vitest
- build: split ecosystem packages

BREAKING CHANGE: VuePress is now published as pure ESM packages
BREAKING CHANGE: CommonJS config file is not supported anymore

- close #617
- close #1008

PR #1049, opened 2022-08-25
chore: remove jest types from typescript config

azat-io:
Jest was removed in https://github.com/vuepress/vuepress-next/pull/1030


meteorlxy:
Uh, thanks!

PR #1052, opened 2022-08-25
refactor(theme-default): set theme data in client config (close #946)

### LINKED ISSUE #946: [Feature request] Client config

Provide client config file for theme and plugins, so that user can modify client-only options to get an HMR instead of a server restart.

### Suggested solution

See https://github.com/vuepress/vuepress-next/issues/914#issuecomment-1143325774

### Alternative

_No response_

### Additional context

_No response_

meteorlxy:
close #946 

BREAKING CHANGE: now you should set default theme locales data in `.vuepress/client.ts` via `defineThemeData`
BREAKING CHANGE: now `plugin-theme-data` does not need options anymore


devPRS:
@meteorlxy Anxiously waiting for this one to come through, will be an amazing change 💯

Mister-Hope:
We shall not remove the old plugin as it's total reasonale to read the options both side.

Some options may related to temp file generating and assets handling.

## Examples

### Tree shakable

A simple example is `editLink` for `@vuepress/theme-default`, overriding it in pages is a rare case, and in most cases we are expecting users to set it globally.

So for a third party theme, it might drop frontmatter support, so that when users set `editLink: false` in options, the theme no longer need to append `filePathRelative` information to page data in `extendsPage` hook.

The above example is simple, but actions like this can  improve "Tree shaking" ability and reduce the final output bundle size. Any global option controlling something all have a chance to do some "tree shakable" work, droping unneeded code.

### Node based options

Some features may related to node based api, and can not be handled at client side, for example, my theme `vuepress-theme-hope` has a feature called "Structured Sidebar", which generate sidebar from file structure automatically. in such cases, users options must be parsed into node side, then passed to client side after edited.

---

So I suggest @meteorlxy reconsidered `plugin-theme-data` rebuilt, old features and behaviors can be (and should be ) preserved when adding the new on.

Mister-Hope:
The problem is that pageRoutes can not be hmred, so if you add new page, a restart is expected.

Radiergummi:
@Mister-Hope I'm kind of facing the same problem, but even more basic. All I want to do is fetch pages from a CMS and add them to the navigation bar, but I can't find a way to modify the navbar programmatically. How did you go about this?

PR #1053, opened 2022-08-25
feat: support layouts option in client config

meteorlxy:
BREAKING CHANGE: theme API `layouts` has been removed, theme layouts should be set in client config instead of theme entry
BREAKING CHANGE: `404` layout should be renamed to `NotFound` layout
BREAKING CHANGE: node API `app.layouts` has been removed


Mister-Hope:
@meteorlxy I found this change awful, with this change, old layouts cannot be tree shaken even if they are overridden to new ones.

PR #1079, opened 2022-09-07
Message for missing content cannot be styled => Add class to div

digital-codes:


Mister-Hope:
I think you can always use a custom NotFound layout. So though I agree that this pr makes some improvements, but I don't think we really need it.

Mister-Hope (approved):

digital-codes:
> I think you can always use a custom NotFound layout. So though I agree that this pr makes some improvements, but I don't think we really need it.
Well, I tried a custom layout at first. But I want the page to look as usual, so basically re-using ParentLayout. Then I cannot get rid of the message "fallback ...", see https://github.com/vuepress/vuepress-next/discussions/1075 
The custom 404 layout sugested in the docs just give a blank page with the 404 message stuff but without navigation and headline. Adding the class seems to be the simplest way to get a regular page with 404 message.

meteorlxy:
In fact, this branch should not be used. If you saw the message, there must be something wrong with your code.

digital-codes:
> 

Well, I copied the default layout because I wanted a full page with navbar, logo and title - plus the 404 message. The example NotFound.vue just gives an empty page with the 404 message. I had another look into the default-theme Layout.vue and copied more stuff to make it work without the "fallback content". You can see the result here: https://vp2.akugel.de/123  (is a 404). The full code for NotFound.vue looks like the following. Pls let me know if I should create a PR for the documentation (not sure yet how to do that though).

```
<script setup lang="ts">
import { useRouteLocale } from '@vuepress/client'
import { useThemeLocaleData } from '@vuepress/theme-default/client'

const routeLocale = useRouteLocale()

const themeLocale  = useThemeLocaleData()

const homeLink = themeLocale.value.home ?? routeLocale.value
const homeText = themeLocale.value.backToHome ?? 'Back to home'
const messages = themeLocale.value.notFound ?? ['Not Found']
const getMsg = (): string =>
  messages[Math.floor(Math.random() * messages.length)]

import Navbar from '@theme/Navbar.vue'


</script>

<template>
  <div
    class="theme-container"
  >
    <slot name="navbar">
      <Navbar>
        <template #before>
          <slot name="navbar-before" />
        </template>
        <template #after>
          <slot name="navbar-after" />
        </template>
      </Navbar>
    </slot>

  <div class="theme-container">
    <main class="page">
      <div class="theme-default-content">
        <h1>404</h1>

        <blockquote>{{ getMsg() }}</blockquote>

        <RouterLink :to="homeLink">{{ homeText }}</RouterLink>
      </div>
    </main>
  </div>

  <div class="my-footer">This is my custom page footer</div>

</div>
</template>


<style lang="css">
.my-footer {
  text-align: center;
}


</style>

<style lang="scss">
/* import default variables from palette */
@import '@vuepress/plugin-palette/palette';

// override  some ...
@import '../../styles/palette.scss';
@import '../../styles/index.scss';


</style>


```

PR #1093, opened 2022-09-18
Preloading critical style-sheets for improved performance 🚤  

sanjaiyan-dev:
This is inspired from `Next.js`.
 1) [Preload critical stylesheets to improve loading speed](https://web.dev/preload-critical-assets/).
 2) Loading CSS top of other prefetches and preloads for faster CSS renders.

_Sorry if I made any mistakes :(_

### DIFF
--- a/ecosystem/theme-default/templates/build.html
+++ b/ecosystem/theme-default/templates/build.html
@@ -23,8 +23,8 @@
 			}
     </script>
     <!--vuepress-ssr-head-->
-    <!--vuepress-ssr-resources-->
     <!--vuepress-ssr-styles-->
+    <!--vuepress-ssr-resources-->
   </head>
   <body>
     <div id="app"><!--vuepress-ssr-app--></div>
--- a/packages/bundler-vite/src/build/renderPageStyles.ts
+++ b/packages/bundler-vite/src/build/renderPageStyles.ts
@@ -12,7 +12,8 @@ export const renderPageStyles = ({
   outputCssAsset: OutputAsset | undefined
 }): string =>
   outputCssAsset
-    ? `
-    <link rel="preload" href="${app.options.base}${outputCssAsset.fileName}" as="style" />
-    <link rel="stylesheet" href="${app.options.base}${outputCssAsset.fileName}" />`
+    ? [
+        `<link rel="preload" href="${app.options.base}${outputCssAsset.fileName}" as="style" />`,
+        `<link rel="stylesheet" href="${app.options.base}${outputCssAsset.fileName}" />`,
+      ].join('')
     : ''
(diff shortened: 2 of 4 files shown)


Mister-Hope:
The first one is useless as it's used in async code. For example you are loading a chunk containing a large package dynamically in a component, and the browser only started to load the resources when the component chunk is loaded and mounted. But a tag like this can tell browsers "Hey, here is a resource we will use later".

But in our case, the stylesheet is loaded immediately, so that line may actually slower the speed.

Mister-Hope:
For the 2 one, I do think we may benifit from this, from my test in the original structure, entrance file `app.js` is loaded before css is actaully loaded, and that may slower our FCP with SSG html.

Mister-Hope:
I personally think that you should revert the first change. Also cc @meteorlxy

Mister-Hope (changes requested):
The first change is not needed.

sanjaiyan-dev:
> The first one is useless as it's used in async code. For example you are loading a chunk containing a large package dynamically in a component, and the browser only started to load the resources when the component chunk is loaded and mounted. But a tag like this can tell browsers "Hey, here is a resource we will use later, download it when you are spare". So when users load and mount the component, the chunk may alredy downloaded, so it can be used at once.
> image
> 
> But in our case, the stylesheet is loaded immediately in next line, so we don't need this.
> 
> 

Sorry, are you saying to remove preloading of critical CSS ?
> Actually I added it because it is recommended by Google web core vitals and it shows a warning to load critical assets ASAP in [Google Web Performance Measuring site](http://web.dev/measure).
And the preload tag won't block the HTML parsing.
And I thought while HTML starts to parse the code the important visual CSS tag is placed below other preload and prefetch tags which can make small visual delay/flicker.


_Extremely sorry if I am wrong :(_

Mister-Hope:
Yes, I don't think that will improve speed.

Mister-Hope:
Yes , because it already block html parsing next line.

If you  move the css and entrance app.js to the bottom of the body, and leave the preload link of css first line then other J's chunks in head, I agree that this does increase speed.

Mister-Hope:
Also, discussions are welcomed here, as I can not garentee my words are all correct either.

Mister-Hope:
I just finished reading related spec , yes, the spec says that it does have benefit.

Mister-Hope (approved):

sanjaiyan-dev:
> I just finished reading related spec , yes, the spec says that it does have benefit.
> 

🤝🙌

sanjaiyan-dev:
> I just finished reading related spec , yes, the spec says that it does have benefit.
> 

@Mister-Hope 

For more ref-: https://web.dev/render-blocking-resources

PR #1094, opened 2022-09-18
perf: render pages in parallel

sanjaiyan-dev:
Improving the performance by running the independent async functions in parallel ✈️.

Ref-: https://github.com/vuejs/vitepress/pull/1320

_Sorry if I made any mistakes :(_

### DIFF
--- a/packages/bundler-vite/src/build/build.ts
+++ b/packages/bundler-vite/src/build/build.ts
@@ -1,6 +1,6 @@
 import type { CreateVueAppFunction } from '@vuepress/client'
 import type { App, Bundler } from '@vuepress/core'
-import { colors, debug, fs, importFile, withSpinner } from '@vuepress/utils'
+import { debug, fs, importFile, withSpinner } from '@vuepress/utils'
 import type { OutputAsset, OutputChunk, RollupOutput } from 'rollup'
 import { build as viteBuild } from 'vite'
 import { resolveViteConfig } from '../resolveViteConfig.js'
@@ -73,20 +73,21 @@ export const build = async (
     const { renderToString } = await import('vue/server-renderer')
 
     // pre-render pages to html files
-    for (const page of app.pages) {
-      if (spinner) spinner.text = `Rendering pages ${colors.magenta(page.path)}`
-      await renderPage({
-        app,
-        page,
-        vueApp,
-        vueRouter,
-        renderToString,
-        ssrTemplate,
-        output: clientOutput.output,
-        outputEntryChunk: clientEntryChunk,
-        outputCssAsset: clientCssAsset,
-      })
-    }
+    await Promise.all(
+      app.pages.map((page) =>
+        renderPage({
+          app,
+          page,
+          vueApp,
+          vueRouter,
+          renderToString,
+          ssrTemplate,
+          output: clientOutput.output,
+          outputEntryChunk: clientEntryChunk,
+          outputCssAsset: clientCssAsset,
+        })
+      )
+    )
   })
 
   // keep the server bundle files in debug mode
--- a/packages/bundler-webpack/src/build/build.ts
+++ b/packages/bundler-webpack/src/build/build.ts
@@ -1,12 +1,6 @@
 import type { CreateVueAppFunction } from '@vuepress/client'
 import type { App, Bundler } from '@vuepress/core'
-import {
-  colors,
-  debug,
-  fs,
-  importFileDefault,
-  withSpinner,
-} from '@vuepress/utils'
+import { debug, fs, importFileDefault, withSpinner } from '@vuepress/utils'
 import webpack from 'webpack'
 import { resolveWebpackConfig } from '../resolveWebpackConfig.js'
 import type { WebpackBundlerOptions } from '../types.js'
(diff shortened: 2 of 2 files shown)


Mister-Hope (changes requested):
This is pretty bad, as it may blow up machine memory. Image my blog have 800+ pages, with each process taking 40MB, I will use 32GB memory.

This may heavily do negative preformance effect due to Node.js "Old Space" behavior, as memory taken greater than old space will be swapped to hardrive.

You may have a test at my blog https://github.com/Mister-Hope/Mister-Hope.github.io, and I am pretty sure you will get the process killed while rendering, and a pages around 200 may get stuck for hours.

You should test this with real project to provide a brechmark with pr like this. The SSG process may actually faster if you make a "Promise queue" and limit it to 5-20 process at same time.

Welcome to try that and show brenchamrk.

sanjaiyan-dev:
> This is pretty bad, as it may blow up machine memory. Image my blog have 800+ pages, with each process taking 40MB, I will use 32GB memory.
> 
> This may heavily do negative preformance effect due to Node.js "Old Space" behavior, as memory taken greater than old space will be swapped to hardrive.
> 
> You may have a test at my blog https://github.com/Mister-Hope/Mister-Hope.github.io, and I am pretty sure you will get the process killed while rendering, and a site with around 200 pages may get stuck for hours.
> 
> You should test this with real project to provide a brechmark with pr like this. The SSG process may actually faster if you make a "Promise queue" and limit it to 5-20 process at same time.
> 
> Welcome to try that and show brenchamrk.

Hi,
Actually I thought node js runs the process in 4 default threads according to [doc](https://nodejs.org/api/cli.html#uv_threadpool_sizesize) and it is also used for I/O tasks. And node js runs tasks concurrently or parallel according to device's capabilities.So I thought it won't break or hang the CPU by overloading because it will run certain tasks concurrently and parallel.

Ref link (about Node js parallel and concurrency) -: https://bytearcher.com/articles/parallel-vs-concurrent


_Just my thoughts might be completely wrong_

sanjaiyan-dev:
> > This is pretty bad, as it may blow up machine memory. Image my blog have 800+ pages, with each process taking 40MB, I will use 32GB memory.
> > 
> > This may heavily do negative preformance effect due to Node.js "Old Space" behavior, as memory taken greater than old space will be swapped to hardrive.
> > 
> > You may have a test at my blog https://github.com/Mister-Hope/Mister-Hope.github.io, and I am pretty sure you will get the process killed while rendering, and a site with around 200 pages may get stuck for hours.
> > 
> > You should test this with real project to provide a brechmark with pr like this. The SSG process may actually faster if you make a "Promise queue" and limit it to 5-20 process at same time.
> > 
> > Welcome to try that and show brenchamrk.
> 
> Hi,
> Actually I thought node js runs the process in 4 default threads according to [doc](https://nodejs.org/api/cli.html#uv_threadpool_sizesize) and it is also used for I/O tasks. And node js runs tasks concurrently or parallel according to device's capabilities.So I thought it won't break or hang the CPU by overloading because it will run certain tasks concurrently and parallel.
> 
> Ref link (about Node js parallel and concurrency) -: https://bytearcher.com/articles/parallel-vs-concurrent
> 
> 
> _Just my thoughts might be completely wrong_

So I thought this won't break the CPU since it only uses the allocated amount of cores. 🥺

Mister-Hope:
I don't think a promise get gc before promise.all is fired. 

However, I will show you some comparisons with my blog with your changes later.

sanjaiyan-dev:
> I don't think a promise get gc before promise.all is fired.
> 
> However, I will show you some comparisons with my blog with your changes later.

Ok, Tks 💪🙌

packages/bundler-webpack/src/build/build.ts
```diff
@@ -95,24 +95,21 @@ export const build = async (
...
-        vueRouter,
-        renderToString,
-        ssrTemplate,
-        allFilesMeta,
-        initialFilesMeta,
-        asyncFilesMeta,
-        moduleFilesMetaMap,
-      })
-    }
+    await Promise.all(app.pages?.map((page) => await renderPage({
```
Mister-Hope, line 101:
```suggestion
    await Promise.all(app.pages?.map((page) => renderPage({
```

Mister-Hope (changes requested):

Mister-Hope (approved):

sanjaiyan-dev:
> I don't think a promise get gc before promise.all is fired. 
> 
> However, I will show you some comparisons with my blog with your changes later.

Hope it worked as expected :)

Mister-Hope:
I already approved this. Meteorlxy is in charge to merge this and publish new version.

sanjaiyan-dev:
> I already approved this. Meteorlxy is in charge to merge this and publish new version.

Sorry, Tks 💪🤝

meteorlxy:
The CI failed

sanjaiyan-dev:
> The CI failed

@meteorlxy Extremely sorry, I think it failed due to some lint issues. 

I used the GitHub web based editor to edit the code can you please help me to solve the issue by running a command:) Extremely sorry for the inconvenience.

packages/bundler-vite/src/build/build.ts
```diff
@@ -72,21 +72,20 @@ export const build = async (
     const { app: vueApp, router: vueRouter } = await createVueApp()
     const { renderToString } = await import('vue/server-renderer')
 
+    if (spinner) spinner.text = `Starting to render pages ${chalk.magenta(app.pages?.map(vuePressPage => vuePressPage.path))}`
```
Mister-Hope, line 75: I dislike the idea here, I would choose to remove page path, for my theme docs around 800 pages, if would be annoying with tons of path throwing to logs without even a line break.

Mister-Hope (changes requested):

packages/bundler-vite/src/build/build.ts
sanjaiyan-dev: > I dislike the idea here, I would choose to remove page path, for my theme docs around 800 pages, if would be annoying with tons of path throwing to logs without even a line break.
```js
if (spinner) spinner.text = `Starting to render pages ${chalk.magenta(app.pages.map(vuePressPage => vuePressPage.path)) + '\n'}`
```

Like this ?

sanjaiyan-dev: \n

sanjaiyan-dev: > I dislike the idea here, I would choose to remove page path, for my theme docs around 800 pages, if would be annoying with tons of path throwing to logs without even a line break.

@Mister-Hope  Sorry, is current change is ok ?

meteorlxy:
Let's just remove the spinner text updates for now.

PR #1100, opened 2022-09-25
feat(bundler-vite): improve chunk name

Mister-Hope:


Mister-Hope:
This ensure a chunk is not started by `_`, solution comming from vitepress, the idea is edit the original `sanitizeFileName` and remove the trailling underscore.

meteorlxy (changes requested):
Fix the CI warning

Mister-Hope:
Done

Mister-Hope:
Edge case: people are using markdown starting with `_`, I will fix this and then mark this ready.

Mister-Hope:
Any update on reviewing this?

Mister-Hope:
Emm that function is redirectly copied from rollup, and was used exactly on the same option.

I though we can trust rollup code as we are using rollup ourselves. however if you do insist a unit test is still needed, I can add some.

meteorlxy:
OK, I forgot that is was forked from rollup. Ignore that.

PR #1101, opened 2022-09-25
feat(bundler-vite): improve chunk split

Mister-Hope:


packages/bundler-vite/src/plugins/mainPlugin.ts
```diff
@@ -4,6 +4,44 @@ import autoprefixer from 'autoprefixer'
 import history from 'connect-history-api-fallback'
 import type { AliasOptions, Connect, Plugin, UserConfig } from 'vite'
 
+const cache = new Map<string, boolean>()
+
+const staticImportedByEntry = (
+  id: string,
+  getModuleInfo: any,
+  cache: Map<string, boolean>,
```
meteorlxy, line 12: The cache arg is unnecessary here

PR #1103, opened 2022-09-25
feat(markdown): support relative image link without `./`

Mister-Hope:
Recognize relative image links not staring with `./` in markdown.

BREAKING CHANGE: markdown image syntax does not support alias anymore, you should use `` tag to use alias

### DIFF
--- a/docs/guide/assets.md
+++ b/docs/guide/assets.md
@@ -8,6 +8,12 @@ You can reference any assets using relative URLs in your Markdown content:
 ![An image](./image.png)
 ```
 
+or
+
+```md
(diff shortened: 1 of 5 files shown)


meteorlxy:
Should also update docs

Mister-Hope:
Done @meteorlxy

Mister-Hope:
I would recommend you to finish docs tweaks and publish this feature in new version. This is stop users switching to vuepress from github wiki or just markdown files in repo.

fdevans:
This breaking change is disappointing and will be a lot of work to upgrade for our site.

Issue #949, opened 2022-06-15
[Feature request] Support customize icon for back-to-top


Mister-Hope:
### Clear and concise description of the problem

The default icon, is emm....

### Suggested solution

Procide a icon option for people to inejct svg strings or html string to customize back-to-top plugin.

PR #1107, opened 2022-09-25
feat(shared): add svg helper

Mister-Hope:


Mister-Hope:
This is for issues like #949. We can make svg customizable.

PR #1111, opened 2022-09-29
build: update shiki

azat-io:
In my project, I use a custom theme for syntax highlighting. I can't update `shiki` version because it is not compatible with the old version


Mister-Hope:
Fix the ci please

Mister-Hope (changes requested):
Issue appears after bumping version

azat-io:
> Issue appears after bumping version

Fixed

PR #1115, opened 2022-10-01
docs: add Russian docs

azat-io:
- [x] translate guide section
- [ ] translate reference section
- [ ] translate advanced section
- [ ] fix links in Russian docs


meteorlxy:
Hi @azat-io , thanks for your interests. However, we are not planing to maintain extra translations in this repo.

If you decide to contribute a Russian translation, you can open another repo for that, and we'll add a link to the Russian site in our official docs.

You can also consider transferring that repo to @vuepress org if it's complete and going to be maintained regularly.

Feel free to share your consideration about that 😄

azat-io:
Why not place documentation in Russian within the main repository? As it happens with Chinese documentation. I could keep the documentation up to date

meteorlxy:
> I could keep the documentation up to date

Appreciate for your interests. In fact, current i18n support of VuePress is not suitable for large-scale projects (like VuePress itself, though). Maintaining multiple translations in a single repo is also not a good practice (see how vuejs does).

Our official docs includes Chinese translation, not only because our core maintainers know Chinese, but also a showcase for the i18n support of VuePress. But as the docs grows, it's really a bad thing you know 😅 

We may migrate the Chinese translations to a standalone repo in the future (and may also migrate docs to a standalone repo, too)

PR #1125, opened 2022-10-10
fix(plugin-search): triggering hotkey while typing

kleinfreund:
## Changes

Fixes an issue with the search plugin’s hotkey being triggered while typing into a text control (e.g. when setting the shortcut to <kbd>/</kbd> and then trying to type `/` into an `input[type=text]`).

## Notes

This added logic is probably not entirely complete and accurate. It would probably be more accurate to only check `input` elements of certain types for example, but I think this should not cause issues.

### DIFF
--- a/ecosystem/plugin-search/src/client/composables/useHotKeys.ts
+++ b/ecosystem/plugin-search/src/client/composables/useHotKeys.ts
@@ -1,8 +1,7 @@
 import { onBeforeUnmount, onMounted } from 'vue'
 import type { Ref } from 'vue'
 import type { HotKeyOptions } from '../../shared/index.js'
-import { isKeyMatched } from '../utils/index.js'
-import { isFocusingTextControl } from '../utils/isFocusingTextControl.js'
+import { isFocusingTextControl, isKeyMatched } from '../utils/index.js'
 
 export const useHotKeys = ({
   input,
--- a/ecosystem/plugin-search/src/client/composables/useHotKeys.ts
+++ b/ecosystem/plugin-search/src/client/composables/useHotKeys.ts
@@ -17,8 +17,8 @@ export const useHotKeys = ({
     if (
       // key matches
       isKeyMatched(event, hotKeys.value) &&
-      // event does not come from search box
-      !input.value.contains(event.target as Node) &&
+      // event does not come from the search box itself or
+      // user isn't focusing (and thus perhaps typing in) a text control
       !isFocusingTextControl(event.target as EventTarget)
     ) {
       event.preventDefault()
--- a/ecosystem/plugin-search/src/client/composables/useHotKeys.ts
+++ b/ecosystem/plugin-search/src/client/composables/useHotKeys.ts
@@ -2,6 +2,7 @@ import { onBeforeUnmount, onMounted } from 'vue'
 import type { Ref } from 'vue'
 import type { HotKeyOptions } from '../../shared/index.js'
 import { isKeyMatched } from '../utils/index.js'
+import { isFocusingTextControl } from '../utils/isFocusingTextControl.js'
 
 export const useHotKeys = ({
   input,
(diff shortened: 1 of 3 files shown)


Mister-Hope:
Actually, the interface for hotKeys is our solution dealing with these situations.

If you have inputs in your site, you probably should not set hotkey with a single key without `ctrl` `alt` or `shift`

kleinfreund:
I don’t think avoiding this sort of conflict should be left to consuming projects. Yes, perhaps single-letter shortcuts aren’t the best idea, but they shouldn’t be ruled out because another tool makes that impossible.

That being said, relying on modifiers to avoid conflicts isn’t as reliable as it seems because several characters are typed while holding down a modifier (most commonly shift). This gets more complex with the wide variety of keyboard layouts in the mix. For example, <kbd>Shift+1</kbd> produces `!` on ANSI US keyboard layouts. Let’s say <kbd>Shift+1</kbd> is used as the search plugin hotkey. The issue will now reproduce when trying to type `!`.

Maybe the majority of such conflicts are very unlikely, but the search plugin could just avoid them by not triggering while the user is typing (which in absence of this particular information would have to be “text control is focused”).

ecosystem/plugin-search/src/client/composables/useHotKeys.ts
```diff
@@ -2,6 +2,7 @@ import { onBeforeUnmount, onMounted } from 'vue'
 import type { Ref } from 'vue'
 import type { HotKeyOptions } from '../../shared/index.js'
 import { isKeyMatched } from '../utils/index.js'
+import { isFocusingTextControl } from '../utils/isFocusingTextControl.js'
```
Mister-Hope, line 5: We prefer to have this import from index.js, a refactor is needed.

ecosystem/plugin-search/src/client/utils/isFocusingTextControl.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Determines whether the user is currently focusing a text control. In this case, the search plugin shouldn’t hijack any hotkeys because the user might be typing into a text field, using type-ahead search in a `select` element, etc.
+ */
+export function isFocusingTextControl(target: EventTarget): boolean {
+  if (target instanceof HTMLElement) {
+    return (
+      document.activeElement === target &&
+      (['TEXTAREA', 'SELECT', 'INPUT'].includes(target.tagName) ||
+        target.hasAttribute('contenteditable'))
+    )
+  } else {
```
Mister-Hope, line 11: Else block for return is not needed

Mister-Hope (changes requested):

ecosystem/plugin-search/src/client/composables/useHotKeys.ts
kleinfreund: I’ve changed this now.

ecosystem/plugin-search/src/client/utils/isFocusingTextControl.ts
kleinfreund: I moved the return statement out of the else block.

Mister-Hope (approved):

ecosystem/plugin-search/src/client/composables/useHotKeys.ts
```diff
@@ -18,7 +18,8 @@ export const useHotKeys = ({
       // key matches
       isKeyMatched(event, hotKeys.value) &&
       // event does not come from search box
-      !input.value.contains(event.target as Node)
+      !input.value.contains(event.target as Node) &&
```
meteorlxy, line 21: Is the condition redundant after adding `isFocusingTextControl` ?

kleinfreund: Yes, it is. The search box (rather what's referenced here as `input.value`) is an input element so it's also covered by `isFocusingTextControl`. I've removed the superfluous condition and left a comment.

PR #1133, opened 2022-10-17
feat(bundler-vite): use utf-8 as encoding, close #1129

### LINKED ISSUE #1129: [Bug report] Outputsize increased by charset
### Description

When built with vite, none ascii characters are formatted to `\uxxx`

image

While actually, giving `charset` to `utf-8` for esbuild, they can be directlly outputed and can help desincrese output size.

### Reproduction

https://v2.vuepress.vuejs.org/

### Used Package Manager

pnpm

### System Info

```shell
Not needed
```

Mister-Hope:


meteorlxy:
Any references for this change?

meteorlxy:
Alright, saw the description of esbuild docs. That makes sense.

PR #1136, opened 2022-10-19
feat(default-theme): Hero image link

rien:
This adds the option `heroImageLink` to the frontmatter which will wrap the hero image in an `<a>` anchor that links to the given URL.

The second commit (d04a0d9fadc1fdf21606a4718e1984b958c4860c) applies this to the docs site, but it can be omitted as it does not add value to the docs site itself.

**Help needed:** Chinese translation of the reference

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,6 +2,7 @@
 home: true
 title: Home
 heroImage: /images/hero.png
+heroImageLink: 'https://v2.vuepress.vuejs.org'
 actions:
   - text: Get Started
     link: /guide/getting-started.html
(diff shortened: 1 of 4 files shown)


Mister-Hope:
I personally don’t think we need this. It’s hard for users to realize logo is clickable, also most of the time, the site is about the logo.

if you do need customization on your site, you are free to use Alia’s to override homepage to customize it yourself

rien:
The `HomeHero`-component which I would need to override has some relative imports, I'm now replacing this with and import from  `@vuepress/theme-default` but it does not seem to find `useDarkMode`.

@Mister-Hope how would you change the imports? I now have:
```ts
import AutoLink from '@theme/AutoLink.vue'
import {
  ClientOnly,
  usePageFrontmatter,
  useSiteLocaleData,
  withBase,
} from '@vuepress/client'
import { isArray } from '@vuepress/shared'
import type { FunctionalComponent } from 'vue'
import { computed, h } from 'vue'
import type { DefaultThemeHomePageFrontmatter } from '@vuepress/theme-default'
import { useDarkMode } from '@vuepress/theme-default'
```
However, this throws `ambiguous indirect export` in the console when trying to run this.

Mister-Hope:
beta52 export everything now

rien:
This works indeed, thanks!

PR #1142, opened 2022-10-23
fix(bundler-vite): fix page routing

### LINKED ISSUE #1140: Error in built docs: ``Referapp.c20e2fa1.js:1 Uncaught ReferenceError: Cannot access 'j' before initialization``
### Description

Hello, 

followed the documentation and started a new vuepress project using ``2.0.0-beta.52`` 
Works fine and great in local dev but built docs throw errors and seems like all js fails because of it (e.g. search, selection of theme light<>dark)

STRP:
1. go to ``src``
2. run ``npm install``
3. run ``npm run docs:build``
4. Upload .vuepress dist folder to some FTP or deploy to github pages
5. observe error in console

image

config:
```ts
import { defineUserConfig, defaultTheme } from 'vuepress'

export default defineUserConfig({
    base: "/wiki-test/",
    lang: 'en-US',
    title: 'Title',
    description: '....',
})
```

### Reproduction

https://github.com/marwie/wiki-test/tree/vuepress/build-error

[dist.zip](https://github.com/vuepress/vuepress-next/files/9843990/dist.zip)

### Used Package Manager

npm

### System Info

```shell
win 10
Chrome Version 106.0.5249.119
```

### Screenshots

app:
image

vendor:
image

azat-io:
Close #1140

### DIFF
--- a/packages/bundler-vite/src/plugins/mainPlugin.ts
+++ b/packages/bundler-vite/src/plugins/mainPlugin.ts
@@ -104,7 +104,8 @@ import '@vuepress/client/app'
                     if (
                       id.includes('plugin-vue:export-helper') ||
                       /@vue\/(runtime|shared|reactivity)/.test(id) ||
-                      /@vuepress\/(client|shared)/.test(id)
+                      /@vuepress\/(client|shared)/.test(id) ||
+                      (/node_modules/.test(id) && /vue/.test(id))
                     ) {
                       return 'framework'
                     }


Mister-Hope (changes requested):
Any packages which is in node_modules and which is being imported twice are supposed to be at `ventor` chunk, and only KNOWN library related to vue and vuepress are splited in to `framework`, this is the exactly same logic with vitepress.

However, we may need to add `vue` and `vue-router` as well, also improve detecting.

You fix might be valid, but it's only goal is to move everything related to framwork. which is not our goal.

Mister-Hope:
We may need to figure out why framework chunk is successfully splitted in the offiical docs, but not in user projects. I would appreciate if you can find issue related to it.

If you just want a quick fix, I would more prefer "revert" the original PR, instead of only making a meaningless workaround here.

meteorlxy:
It's obvious that in our repo, the `@vuepress\/client`, etc. won't be matched, because the final resolved path is not in the `node_modules.`

Mister-Hope:
@meteorlxy My pr only match package name, and it is actually matching, so official docs has framework chunk, however this is not happening in user projects, pretty strange.

meteorlxy:
Closing as we'll revert #1101 for now

PR #1143, opened 2022-10-24
feat(cli): use help as default command

azat-io:
image


meteorlxy:
Thanks for raising this feature. Implemented in bcc2208

PR #1145, opened 2022-10-25
feat(utils): migrate to picocolors

azat-io:
`picocolors` 14 times smaller and 2 times faster than `chalk`.

VitePress also use `picocolors`:
https://github.com/vuejs/vitepress/pull/659


meteorlxy (changes requested):
Should use a shorter name or even replace `chalk` directly

Mister-Hope:
Just as a reminder: This will be a breaking change. As most of my plugins are using chalk imported from `@vuepress/utils` to print logs in terminal.

azat-io:
> I think we should re-export `picocolors` to something shorter (like `colors` or even `chalk`)

Renamed to `colors`

> Just as a reminder: This will be a breaking change. As most of my plugins are using chalk imported from `@vuepress/utils` to print logs in terminal.
> 
> So AT LEAST it should NOT be marked as `chore`

Updated PR type to `feat`

meteorlxy (approved):

PR #1158, opened 2022-11-09
feat(theme-default): collapsible nested item in sidebar (close #883)

### LINKED ISSUE #883: [Feature request] Officially support `collapsible: true` deeper than top level items

Currently, `collapsible: true` _can_ be set on nested sidebar items and it _does_ work, but the Typescript typings don't allow this so I have to cast to `any`.

The functionality today works how I would want it - the nested items under Attributes in my example below are listed when navigated into that section, and not listed when not in that section. 

It is just that the documentation and the TS types try to discourage this usage. It would also be nice if the arrow to manually expand/collapse without navigating into the section could be rendered.

### Suggested solution

Allow this to be an officially supported scenario without the cast to `any` on the child item:

``` ts
sidebar: [
  "/",
  {
    text: 'Model Components',
    children: [
      '/model-components/properties',
      {
        text: "Attributes",
        link: '/model-components/attributes',
        collapsible: true, // Allow this to work without TS workarounds.
        children: fs
          .readdirSync(path.resolve(__dirname, '../model-components/attributes'))
          .map(f => '/model-components/attributes/' + f)
      } as any,
      '/modeling/model-components/methods',
     ],
  },
] 
```

### Alternative

_No response_

### Additional context

_No response_

pplr:
close #883

### DIFF
--- a/docs/reference/default-theme/config.md
+++ b/docs/reference/default-theme/config.md
@@ -313,7 +313,7 @@ export default {
 
   To configure the sidebar items manually, you can set this option to a _sidebar array_, each item of which could be a `SidebarItem` object or a string:
 
-  - A `SidebarItem` object should have a `text` field, could have an optional `link` field and an optional `children` field. The `children` field should be a _sidebar array_. When a `SidebarItem` object is placed at the root-level, it could have an extra optional `collapsible` field to control whether it is collapsible.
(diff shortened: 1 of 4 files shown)


meteorlxy:
Sorry for delay. 

Thanks for your contributions! ❤️

pplr:
Sorry for being impatient! Thanks

PR #1170, opened 2022-11-23
feat(bundler-vite): improve chunk split

Mister-Hope:


Mister-Hope:
This should work, I found the reason why it breaks in user env.

@vuepress/client generate temp files for clientConfigs, and it includes all client entry of other plugins and theme, so rollup will think those shall all included in ventor chunk.

If we only split Vue, VueRouter and @vuepress/shared to framework, everything will be fine, and we split out a 80kb chunk to avoid update per deploy.

I try to read from `app.dir.temp()` and omit those `internal` files output @vuepress/client, but in that casae @vuepress/client becomes very "slim", and I personally think it's not worth spliting it in that case. Our main goal is to split out vue, vue-router and plugin:vite-helper as the first and second is large in app.js and the third one can actauly increase the final size if it's generated as a chunk by rollup

PR #1176, opened 2022-12-01
fix(client): fix double quote in head tag

### LINKED ISSUE #1174: [Bug report] doule quote in head tag content cause issues in devServer
### Description

```md
---
head:
  - - meta
    - name: test
      content: 'some text with "double quote"'

---

Content
```

image

The css selector is not valid anymore under this situation.

tinyfind:
fixed #1174 https://github.com/vuepress-theme-hope/vuepress-theme-hope/issues/2448#issue-1461331094

### DIFF
--- a/packages/client/src/setupUpdateHead.ts
+++ b/packages/client/src/setupUpdateHead.ts
@@ -85,7 +85,7 @@ export const queryHeadTag = ([
   const attrsSelector = Object.entries(attrs)
     .map(([key, value]) => {
       if (isString(value)) {
-        return `[${key}="${value}"]`
+        return `[${key}=${JSON.stringify(value)}]`
       }
       if (value === true) {
         return `[${key}]`


Mister-Hope:
You’d better explain why JSON.stringing is needed.

Mister-Hope:
Closing as this is not a valid fix, if attr contain space, you fix can cause similar issue.

tinyfind:
> Closing as this is not a valid fix, if attr contain space, you fix can cause similar issue.

sure?  can you give me an example

tinyfind:
> You’d better explain why JSON.stringify is needed.

> `'some text with "double quote"'` valid string
https://github.com/vuepress/vuepress-next/blob/8c345d827c758326429b20f37b6be95f0865c655/packages/client/src/setupUpdateHead.ts#L88
but here will transform that to "`some text with "double quote""`，it's invalid
so use  `""` or `''`  will generate invalid string,

Mister-Hope:
```html
<meta name="description" value="Some words here" />
```

Mister-Hope (approved):

PR #1181, opened 2022-12-05
feat: improve print style

Mister-Hope:


Mister-Hope:
Before:
image

After:

image

meteorlxy (changes requested):
I think we should put all print-related styles together in a same file or in a same section of their corresponding style file.

What about just adding a `print.scss` file in default theme? This might be more related to theme than those plugins.

PR #1208, opened 2022-12-26
feat(plugin-docsearch): add injectStyles option

Mister-Hope:


Mister-Hope:
I am not sure about the option name, maybe we can choose better one.

meteorlxy (approved):

PR #1220, opened 2022-12-29
fix(markdown): respect token attrs in code block

Mister-Hope:


Mister-Hope:
Some 3rd party markdown-it plugin about code blocks is effected with this bug.

meteorlxy (approved):

PR #1224, opened 2023-01-05
feat(plugin-docsearch): support indexBase option (close #1223)

### LINKED ISSUE #1223: [Bug report] Docsearch result incorrect
### Description

When deploying docs at mutiplce places with different base, the other site cannot gent a search result correctly.

For example, vuepress-theme-hope has documentation on:

- https://vuepress-theme-hope.github.io/v2/
- https://theme-hope.vuejs.press/

And currently docsearch is crawling the fiirst one, however, a `https://vuepress-theme-hope.github.io/v2/test.html` will be converted to a routerlink `/v2/test.html`

This will happens as long as docs are deployed at different base. 

### Solution

We may need a setting called `indexBase` so that when coverting to internal route link, the base can be removed correctly.

Mister-Hope:


meteorlxy:
Then do you think we should support different `indexBase` in different locales?

Mister-Hope:
> Do you think we need to support different `indexBase` in different locales?

Maybe, I am not sure.

Since our locales is based on  file structure and a bit hard to customize (there are hard links in serverl pages like homepage frontmatter `actions` option), I am not expecting users to reorganize files with scripts or use permalinks to override them and deal all hard links.

For my users cases, they are not changing locales.

meteorlxy:
OK, let's keep it simple to meet most use cases. I've updated the docs and you could check it again.

Mister-Hope:
LGTM

meteorlxy (approved):

PR #1229, opened 2023-01-07
refactor: use nullish coalescing assignment

Mister-Hope:


Mister-Hope:
Emm, the test failed because Node 14 do not support `??=`, however, the output code by tsup does not contain `??=` either.

If we do want to add tests for node 14 and node 16 besides the development env node18. we'd better use output file for testing.

@meteorlxy Is the above sounds good to you?

meteorlxy:
Yes that's reasonable. We should make sure tsup is generating node 14 compatible code, too.

PR #1230, opened 2023-01-07
feat: add helper package

Mister-Hope:
These are common packages which can benifit plugin and theme developers.

Some docs could be found at:

- https://shared.vuejs.press/node/bundler.html
- https://shared.vuejs.press/node/content.html

However, I would recommand to add this to vuepress-next repo. I am not sure whether you prefer adding docs on main docs, or make it standalone.

I am actually planing to write some cookbooks about how to write plugins and themes at vuejs.press


docs/reference/plugin-api.md
```diff
@@ -169,29 +169,57 @@ export default {
...
+
+    // Modify webpack options through chainWebpack, e.g.: add a webpack plugin
+    addChainWebpack(bundlerOptions, app, (config) => {
+      config.plugin('your-plugin').use(YourPlugin, yourPluginOptions)
+    })
+
+    // Modify vite config by merging your options, e.g.: add a vite plugin
+    addViteConfig(bundlerOptions, app, {
+      plugins: [YourVitePlugin(yourVitePluginOptions)],
+    })
```
Mister-Hope, line 211: I think these functions are definitely useful to plugin or theme developers, as they no longer need to write long codes.

The original code block has 20 lines code to achieve the following one:

```ts
addCustomElement(bundlerOptions, app, 'my-custom-element')
```

ecosystem/helper/src/client/utils/path.ts
```diff
@@ -0,0 +1,47 @@
...
+    .replace(/(index)?\.(md|html)$/, '')
+
+/**
+ * Whether the lick is active
+ *
+ * @param route Current route
+ * @param link link path
+ * @returns Whether the lick is active
+ */
+export const isActiveLink = (
```
Mister-Hope, line 31: This will be useful in many places, such as catalog, toc component or any comopnent which has active state with links. so this is extracted from theme-default

ecosystem/helper/src/client/utils/resolveRouteWithRedirect.ts
```diff
@@ -0,0 +1,28 @@
+import { isFunction, isString } from '@vuepress/shared'
+import type { Router } from 'vue-router'
+
+/**
+ * Resolve a route with redirection
+ */
+export const resolveRouteWithRedirect = (
```
Mister-Hope, line 7: so this is extracted from theme-default

ecosystem/helper/src/client/utils/hasGlobalComponent.ts
```diff
@@ -0,0 +1,14 @@
...
+export const hasGlobalComponent = (name: string, app?: App): boolean => {
+  const instance = app ? app._instance : getCurrentInstance()
+
+  return (
+    isPlainObject(instance?.appContext.components) &&
+    (name in instance!.appContext.components ||
+      camelize(name) in instance!.appContext.components ||
+      capitalize(camelize(name)) in instance!.appContext.components)
+  )
+}
```
Mister-Hope, line 14: This helper function should be useful enough, as plugins and themes should check if there is already same name global component when registing one to prevent Vue displaying warnings.

ecosystem/plugin-external-link-icon/src/client/config.ts
```diff
@@ -9,7 +10,11 @@ const locales = __EXTERNAL_LINK_ICON_LOCALES__
...
 export default defineClientConfig({
   enhance({ app }) {
-    // wrap the `<ExternalLinkIcon />` component with plugin options
-    app.component('ExternalLinkIcon', h(ExternalLinkIcon, { locales }))
+    // only register when ExternalLinkIcon is not found in global components
+    // so that users can provide their own icon component
+    if (hasGlobalComponent('ExternalLinkIcon', app)) {
+      // wrap the `<ExternalLinkIcon />` component with plugin options
+      app.component('ExternalLinkIcon', h(ExternalLinkIcon, { locales }))
+    }
```
Mister-Hope, line 18: This is more likely to be an example why we need `hasGlobalComponent`, if you prefer to have this change in a sperate pr, I can revert it and open another one once you merged this pr.

Mister-Hope:
Some explaination why we need these code are added.

ecosystem/helper/src/node/bundler/vite/mergeViteConfig.ts
```diff
@@ -0,0 +1,188 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Mister-Hope, line 1: This file is absolutely needed, as some developers will choose to achieve config merging by import this function from 'vite', however, no matter how slim vite it is, it's still not a small package, so extracting `mergeViteConfig` and provide it standalone should benifit those who choose `bundler-webpack`

ecosystem/helper/src/node/bundler/customElement.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    if (!viteBundlerConfig.vuePluginOptions.template)
+      viteBundlerConfig.vuePluginOptions.template = {}
+
+    if (!viteBundlerConfig.vuePluginOptions.template.compilerOptions)
+      viteBundlerConfig.vuePluginOptions.template.compilerOptions = {}
+
+    const {
+      isCustomElement = (tag: string): void => tagHint(tag, app.env.isDebug),
+    } = viteBundlerConfig.vuePluginOptions.template.compilerOptions
```
Mister-Hope, line 56: Could be changed to the following, however I am not sure if you prefer to have dev env run at node14, support is starting from node16.


```suggestion
    const {
      isCustomElement = (tag: string): void => tagHint(tag, app.env.isDebug),
    } = (((viteBundlerConfig.vuePluginOptions ??= {}).template ??=
      {}).compilerOptions ??= {});
```

meteorlxy:
Closing as we have migrated ecosystem into [another repo](https://github.com/vuepress/ecosystem)

Issue #123, opened 2021-04-25
[Bug report] Page flickers after initial load


KnorpelSenf:
### Description

When the page loads, it first displays the frontpage, then switches to an all white screen, and then displays it correctly again. This flickering takes less than a second in both Chrome and Firefox with no cache, and gets faster on subsequent page reloads. This affects only the front page.

### Steps to reproduce

- Reproduction link / repo:

  Look at <https://grammy.dev/> which is built from [here](https://github.com/grammyjs/website/tree/main/site) using `"vuepress": "^2.0.0-beta.9"`.

In fact, you can also observe it on the documentation of Vuepress Next itself at <https://vuepress2.netlify.app/> if you throttle your network in the devtools, e.g. to Slow 3G.

This does not happen for Vuepress itself, <https://vuepress.vuejs.org/> works just fine under all circumstances.

### Expected behavior

When the page loads, the content does not disappear again after being displayed.

### Screenshots

![rec-2021-04-25_13 52 14 (1)](https://user-images.githubusercontent.com/12952387/115992956-9d8d5d00-a5d0-11eb-9df6-b09690f2acb3.gif)

### Environment info

- Browser: Chrome (Linux), Chrome (Android), Firefox ESR (Linux)
- Output of `vuepress info`:

```bash
$ npx vuepress info

  System:
    OS: Linux 4.19 Debian GNU/Linux 10 (buster) 10 (buster)
    CPU: (4) x64 Intel(R) Core(TM) i5-4200M CPU @ 2.50GHz
    Memory: 2.59 GB / 7.55 GB
    Shell: 5.0.3 - /bin/bash
  Binaries:
    Node: 14.16.1 - /usr/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.6.1 - /usr/local/bin/npm
  Utilities:
    Git: 2.20.1 - /usr/bin/git
  Browsers:
    Chrome: 90.0.4430.85
    Firefox: 78.10.0esr
  npmPackages:
    @vuepress/bundler-vite: Not Found
    @vuepress/bundler-webpack:  2.0.0-beta.9 
    @vuepress/cli:  2.0.0-beta.9 
    @vuepress/client:  2.0.0-beta.9 
    @vuepress/core:  2.0.0-beta.9 
    @vuepress/markdown:  2.0.0-beta.9 
    @vuepress/plugin-active-header-links:  2.0.0-beta.9 
[...]

PR #1237, opened 2023-01-13
Docs: Update example file extension to ts

laurensgroeneveld:

- [ ] Bug fix
- [ ] New feature
- [X] Documentation update
- [ ] Other

### Description

Tiny fix: All files on this documentation page are Typescript files, but the example client config for the child theme is `client.js` instead of `client.ts`, even though the example client config above it is also a TS file.

### DIFF
--- a/docs/reference/default-theme/extending.md
+++ b/docs/reference/default-theme/extending.md
@@ -100,7 +100,7 @@ export const childTheme = (options: DefaultThemeOptions): Theme => {
     extends: defaultTheme(options),
 
     // override layouts in child theme's client config file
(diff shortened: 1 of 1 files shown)


meteorlxy:
Hi, thanks for your contribution.

For this part, it was intentionally written as `.js` extension. As a theme author, you are likely to publish your theme to npm. Before publishing, you'd better transpile the ts source code into js code, then the filename should be `.js`.

Maybe we should add some description here.

laurensgroeneveld:
Ah, that makes sense 👍 I was working on child theme for the default theme so had already copied the `client.ts` example. A comment explaining why it is `.js` here would be helpful!

PR #1246, opened 2023-01-29
ci: simplify release github workflow

azat-io:


Mister-Hope:
So what's rhe difference here

azat-io:
- make it easier to move the repository if it happens
- avoid mistakes in case of copy-paste workflow code
- variables are always better and cleaner

PR #1253, opened 2023-02-05
fix(client): avoid updating routeLocale

Mister-Hope:


packages/client/src/setupGlobalComputed.ts
```diff
@@ -87,6 +86,16 @@ export const setupGlobalComputed = (
     resolvers.resolvePageLayout(pageData.value, layouts.value)
   )
 
+  // router.currentRoute is a ref, so we must watch routePath explicitly
+  watch(
+    () => router.currentRoute.value.path,
+    (newValue) => {
+      if (newValue !== routePath.value) {
```
meteorlxy, line 93: Is the `if` branch necessary?

BTW, could you check if the `computedWithControl` from vueuse is helpful for this case?

Mister-Hope: >Is the if branch necessary?

It's not, removed

> BTW, could you check if the computedWithControl from vueuse is helpful for this case?

Tried, but seems not working, seems like the same as the original, as long as route object change, computed is triggered.

I think that using `computedWithControl ` should be the same as

```
const routePath = computed(()=>  router.currentRoute.value.path)
```

routePath  and everything depending on it will still be triggered while scrolling.

meteorlxy: I just checked the source code of `computedWithControl`. It's using `watch` inside. 🤔 

How did you try it? Something like this? (Just curious and discuss)

```ts
const routePath = computedWithControl(
  ()=>  router.currentRoute.value.path,
  ()=>  router.currentRoute.value.path
)
```

meteorlxy (approved):

PR #1260, opened 2023-02-15
fix(theme-default): fix inactive style

ntnyq:

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

### Description

Text overflow ellipsis is inactive because `span.site-name` is an inline element.

The style `width` takes no effect.

The rule `.can-hide` is moved to ensure it can overrides `display: block;` when there is a navbarBrandLogo.

The `width` is changed to adapt the situation that there is both a ColorModeSwitch and a NavbarSearch.

### Screenshots

**Before**

Screenshot 2023-02-15 at 17 24 21

**After**

Screenshot 2023-02-15 at 17 37 49

### DIFF
--- a/ecosystem/theme-default/src/client/styles/navbar.scss
+++ b/ecosystem/theme-default/src/client/styles/navbar.scss
@@ -43,15 +43,16 @@
   .navbar {
     padding-left: 4rem;
 
-    .can-hide {
-      display: none;
-    }
-
     .site-name {
-      width: calc(100vw - 9.4rem);
+      display: block;
+      width: calc(100vw - 11rem);
       overflow: hidden;
       white-space: nowrap;
       text-overflow: ellipsis;
+    }
+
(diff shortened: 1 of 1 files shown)


ecosystem/theme-default/src/client/styles/navbar.scss
```diff
@@ -43,16 +43,17 @@
...
     padding-left: 4rem;
 
-    .can-hide {
-      display: none;
-    }
-
     .site-name {
-      width: calc(100vw - 9.4rem);
+      display: block;
+      width: calc(100vw - 11rem);
```
Mister-Hope, line 48: Could you add comments about 11rem?

ntnyq: Comment added.

Mister-Hope (approved):

PR #1269, opened 2023-02-21
chore: tweaks

Mister-Hope:


meteorlxy:
It's OK to explicitly define the `required` option

Mister-Hope:
But the change can reduce size, also eslint-plugin-vue has similar rules that disallow required: false and any behaviour which is not the same as `prop: Boolean` (default behaviour for Boolean props on custom element)

Mister-Hope:
Hello? Maybe a relook into this one?

Mister-Hope:
> @Mister-Hope - IMO, source code readability makes more sense than such kind of "size reduction".

I still do not get why there is any readability improvement here.

https://vuejs.org/guide/components/props.html#prop-validation

The official docs never shows an example with both required:false and default. And it clearly says that only required:true is meaningful.

And here are contents from Selina-plugin-vue:

```vue/no-required-prop-with-default #
Enforce props with default values to be optional
🔧 The --fix option on the command line can automatically fix some of the problems reported by this rule.
💡 Some problems reported by this rule are manually fixable by editor suggestions.
📖 Rule Details #

If a prop is declared with a default value, whether it is required or not, we can always skip it in actual use. In that situation, the default value would be applied. So, a required prop with a default value is essentially the same as an optional prop. This rule enforces all props with default values to be optional.
```


So why is a useless declaration increase readability?

PR #1272, opened 2023-02-27
feat(markdown): add error logger for importCode plugin

Mister-Hope:


packages/markdown/src/plugins/importCodePlugin/resolveImportCode.ts
```diff
@@ -15,6 +15,7 @@ export const resolveImportCode = (
     // if the importPath is relative path, we need to resolve it
     // according to the markdown filePath
     if (!filePath) {
+      console.error(`Import file ${importPath} can not be resolved`)
```
meteorlxy, line 18: Use `logger` from utils

meteorlxy (changes requested):

Mister-Hope:
Done

PR #1285, opened 2023-03-17
chore: bump deps

Mister-Hope:


Mister-Hope:
Tsup not yet support typescript v5, so we shall wait for it.

PR #1286, opened 2023-03-19
feat: add support for clientData

Mister-Hope:


Mister-Hope:
@meteorlxy  I build a feature called client data, and you can check this.

Docsearch plugin is modified as an example.

packages/client/src/helpers/defineClientData.ts
```diff
@@ -0,0 +1,13 @@
+import type { InjectionKey } from 'vue'
+
+export const store = new Map<InjectionKey<unknown>, unknown>()
```
meteorlxy, line 3: Rename to something like `clientDataMap` to be more specific?

Mister-Hope:
@meteorlxy This pr just include clientData api itself now, if you think that the api is ready, we can release this in next version so that I can start to use it and collect some feedbacks from my users.

meteorlxy:
@Mister-Hope In fact, I don't see too many differences from provide/inject:

```ts
export default defineClientConfig({
  enhance({ app }) {
    app.provide(foo, bar)
  },
})
```

Any further ideas?

Mister-Hope:
> @Mister-Hope In fact, I don't see too many differences from provide/inject:
> 
> ```ts
> export default defineClientConfig({
>   enhance({ app }) {
>     app.provide(foo, bar)
>   },
> })
> ```
> 
> Any further ideas?

This ensures that the clientData can be visited in any enhance function.

There will be cases where a plugin is providing some exports which requires client data

Mister-Hope:
Any update on this?


> This ensures that the clientData can be visited in any enhance function.
> 
> There will be cases where a plugin is providing some exports which requires client data.
> 
> I make a draft with the code you wrote in my theme, and I met similar problems. The theme clientConfig seems not be the first one, so I cannot access the data in enhance cycle in some cases.
> 
> Moving it to core insure every data is injected before any client hook.

I think this reason is enough to add this feature in core.

Mister-Hope:
cc again @meteorlxy

meteorlxy:
A `beforeEnhance` option might be more general?

Mister-Hope:
I think a `beforeEnhance` option is pretty weird, as I am not thinking out any other things this option can solve.

Let's focus on the problem of this pr:

- User should be allowed to set options in client side
- All plugins should be able to read these client configs, so that users do not need to set repeatly with these configs
- The currently implementation is that a plugin manually `provide` its own config within enhance lifecycle, so that the plugin sequence decides the client config scope where a plugin or theme could access

Adding `beforeEnhance` could be a solution, but if there is no other use for it, a normal developer or users which does not need a client config could be confused about the hook.

Adding it to core as a new concept seems more natural to me.

Another improvement is that we now requires users to add a  `export default {}`.

We could also allow the user client config to do not have any default export, i.e.:

```ts
// .vuepress/client.ts
// imports

definePluginAConfig({ xxx })
```

Mister-Hope:
> Another improvement is that we now requires users to add a `export default {}`.
> 
> We could also allow the user client config to do not have any default export, i.e.:
> 
> ```ts
> // .vuepress/client.ts
> // imports
> 
> definePluginAConfig({ xxx })
> ```

I would like to open a pr for this, this is useful for users that only want to add this file for plugin customization purpose, it will be a bit strange for them to understand that without a default export we are throwing an error.

Also, A lot of plugins are not expecting usage with the export object, maybe they just want to add some styles or library imports on the client side.

PR #1308, opened 2023-04-20
fix(core): ensure page.title is string, close #1306

### LINKED ISSUE #1306: [Bug report] Page.title is not type guarded
### Description

```md
---
title: 1
---
```

```md
---
title: {}
```

There will be noob users who set titles like this, and `page.title` (with `string` type) will have other types, and this may break the application.

Unless we set `page.title: unknown` , we should guard the types of it.

### Reproduction

Not needed

### Used Package Manager

npm

### System Info

```shell
.
```

Mister-Hope:


packages/core/src/page/renderPageContent.ts
```diff
@@ -74,6 +74,6 @@ export const renderPageContent = ({
       'frontmatter'
     ),
     sfcBlocks,
-    title: frontmatter.title ?? title,
+    title: isString(frontmatter.title) ? frontmatter.title : title,
```
meteorlxy, line 77: Maybe

```
frontmatter.title ? `${frontmatter.title}` : title
```

Mister-Hope: Emm, then why not `String(fromtmatter.title)`

meteorlxy: As I remember, string template has better performance than string constructor in most cases.

The only difference is that string constructor can handle symbol, while string template cannot. But I think frontmatter.title is impossible to be a symbol value in this case.

Mister-Hope, line 77:
```suggestion
    title: frontmatter.title ? `${frontmatter.title}` : title,
```

Mister-Hope:
This should be ready

PR #1314, opened 2023-04-20
fix(client): avoid updating existing head tags, close #1268

### LINKED ISSUE #1268: [Bug report] Head resources are loaded again when switching pages
### Description

When users set js and css resources through head option in config file, switching pages will cause these resources to relead.

I think this should be bug as we should diff the head list and try to avoid upgrading the unchanged one

### Reproduction

https://dromara.org/fast-request/

### Used Package Manager

npm

### System Info

```shell
Not related
```

Mister-Hope:


Mister-Hope:
@meteorlxy Need some help here, the id is recalculated many times, and I am thinking if we can add a new type called `ResolvedHeadConfig` besides `HeadConfig` while has the id attribute?

jcwangxp:
@meteorlxy the js reload issue causes wired behavior, and we need this bug to be fixed, so could you help to review and get it fixed?

meteorlxy:
Some alternatives:
- Make site data head config more stable - only updated according to route locale
- Inject site data head config into html template

Mister-Hope:
I am actually "upvote" for this kind of changes, it's common to merge a pr and create a new commit through things like github, and people can choose to hide their emails:

image

It's definitely better if we can recognize them in plugin.

meteorlxy:
ref: https://github.com/vuejs/vitepress/pull/3017

Mister-Hope:
@meteorlxy Maybe a review of this pr?

meteorlxy:
@Mister-Hope e2e failed for some reason. Also meet the error when testing locally. Try `cd e2e && pnpm e2e:dev`

Mister-Hope:
I give up this pr, the problem is that in production, head tag already exisits in `<head>`, while on devServer it's not, I can not find a graceful way to bypass this right now.

PR #1315, opened 2023-04-21
perf(shared): avoid using regexp match

Mister-Hope:


meteorlxy:
Could you quantify how much performance does it improve?

BTW, how much does it benefit by using `str[0] === '/'` over `str.startsWith('/')` ? `startsWith`/`endsWith` should have better readability.

Mister-Hope:
> Could you quantify how much performance does it improve?
> 
> BTW, how much does it benefit by using `str[0] === '/'` over `str.startsWith('/')` ? `startsWith`/`endsWith` should have better readability.

20%, the latter is much more greater than 20

Mister-Hope:
Any update on this?

meteorlxy (approved):

PR #1322, opened 2023-04-28
feat(theme-default): optimizing css variable acquisition

zhaopan-pan:
optimize the acquisition method of css variables and encapsulate related methods


ecosystem/theme-default/src/client/utils/updateDeviceStatus.ts
```diff
@@ -0,0 +1,27 @@
...
+const DeviceTypeMap = {
+  [DeviceType.MOBILE]: Number(cssVars.mobile?.replace('px', '')),
+}
+
+/**
+ * add listener to detect screen though device type
+ */
+export const updateDeviceStatus = (
+  deviceType: DeviceType,
+  callback: (w: number) => void
```
Mister-Hope, line 17:
```suggestion
  callback: (width: number) => void
```

Mister-Hope (approved):

zhaopan-pan:
The above checks error seems to be caused by the css-loader of webpack not opening cssMoudle🤔

ecosystem/theme-default/src/client/shim.d.ts
```diff
@@ -3,3 +3,11 @@ declare module '*.vue' {
   const comp: ComponentOptions
   export default comp
 }
+
+declare module '*.module.scss?module' {
+  type Variables = {
+    [className: string]: string
+  }
+  const cssVar: Variables
+  export default cssVar
```
meteorlxy, line 12:
```suggestion
  const cssVars: Record<string, string>
  export default cssVars
```

ecosystem/theme-default/src/client/utils/updateDeviceStatus.ts
```diff
@@ -0,0 +1,27 @@
+import { onMounted } from 'vue'
+import cssVars from '../styles/_variables.module.scss?module'
+
+export enum DeviceType {
+  MOBILE = 'mobile',
+}
+
+const DeviceTypeMap = {
+  [DeviceType.MOBILE]: Number(cssVars.mobile?.replace('px', '')),
```
meteorlxy, line 9: `Number.parseInt()` could be better

ecosystem/theme-default/src/client/utils/updateDeviceStatus.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * add listener to detect screen though device type
+ */
+export const updateDeviceStatus = (
+  deviceType: DeviceType,
+  callback: (width: number) => void
+): void => {
+  const width = DeviceTypeMap[deviceType]
+  if (!width) return
```
meteorlxy, line 20:
```suggestion
  if (!Number.isInteger(width)) {
      if (__DEV__) throw new Error('xxxx')
      return
  }
```

meteorlxy (commented):
Nice implementation! Suggested some minor changes

ecosystem/theme-default/src/client/shim.d.ts
zhaopan-pan: Here is an easier way, I optimized it

ecosystem/theme-default/src/client/utils/updateDeviceStatus.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const DeviceTypeMap = {
+  [DeviceType.MOBILE]: Number.parseInt(cssVars.mobile?.replace('px', ''), 10),
+}
+
+/**
+ * add listener to detect screen though device type
+ */
+export const updateDeviceStatus = (
```
meteorlxy, line 15: This method is more like a composable than a util. It's required to be used in component setup. So better to move it under composables dir and rename it with `use` prefix

ecosystem/theme-default/src/client/components/Navbar.vue
```diff
@@ -24,27 +25,20 @@ const linksWrapperStyle = computed(() => {
...
+      navbar.value!.offsetWidth -
+      navbarHorizontalPadding -
+      (navbarBrand.value?.offsetWidth || 0)
   }
-  handleLinksWrapWidth()
-  window.addEventListener('resize', handleLinksWrapWidth, false)
-  window.addEventListener('orientationchange', handleLinksWrapWidth, false)
-})
+}
+updateDeviceStatus(DeviceType.MOBILE, handleLinksWrapWidth)
```
meteorlxy, line 41: Maybe simplify to `updateDeviceStatus(DeviceType.MOBILE, (...) => { ... })`

Seems that we don't need to declare the callback function separately

meteorlxy:
@zhaopan-pan added 2 more comments

ecosystem/theme-default/src/client/utils/updateDeviceStatus.ts
zhaopan-pan: I added it under composites

zhaopan-pan:
> @zhaopan-pan added 2 more comments

i has optimized it

meteorlxy (approved):

meteorlxy:
Thanks for your contributions! ❤️

PR #1324, opened 2023-05-06
Fixing Document Errors

MoonLord-LM:
【Fixing Document Errors】
There are omissions and inconsistencies in the document.
According to the current npm document, the project cannot be started at all.

【修复文档错误】
文档内容存在遗漏和不一致
按照当前的 npm 文档操作，工程根本启动不起来


Mister-Hope:
npm does not require strict deps in my mind, so does yarn classic(v1)

MoonLord-LM:
I don't know if it's my local npm cache, but I tried it again and it was fine.

不知道是不是我本地 npm 缓存的问题，重新试了结果又没问题了 = =

PR #1327, opened 2023-05-11
feat: bump vue to 3.3

Mister-Hope:


Mister-Hope:
Note: eslint-plugin-vue does not add `defineSlots` maros yet, so there is a workaround to hide no-def error until new version that supports it.

ecosystem/theme-default/src/client/components/Navbar.vue
```diff
@@ -10,7 +10,13 @@ import {
   useUpdateDeviceStatus,
 } from '../composables/index.js'
 
-defineEmits(['toggle-sidebar'])
+defineEmits<{
+  (e: 'toggle-sidebar'): void
+}>()
+defineSlots<{
```
meteorlxy, line 16: Add an empty line before

ecosystem/theme-default/src/client/components/Page.vue
```diff
@@ -1,6 +1,13 @@
 <script setup lang="ts">
 import PageMeta from '@theme/PageMeta.vue'
 import PageNav from '@theme/PageNav.vue'
+
+defineSlots<{
+  'top'?: (props: Record<never, never>) => any
```
meteorlxy, line 6: Any way to avoid those verbose `Record<never, never>` and `any` ?

.eslintrc.cjs
```diff
@@ -2,6 +2,8 @@ module.exports = {
   root: true,
   extends: 'vuepress',
   globals: {
+    // workaround for vue3.3 slots
+    defineSlots: 'readonly',
```
meteorlxy, line 6: Maybe put after `__` properties

ecosystem/theme-default/src/client/components/Page.vue
Mister-Hope: Currently volar complains if props are not provided, the any is expected.

See https://blog.vuejs.org/posts/vue-3-3#typed-slots-with-defineslots and https://github.com/vuejs/core/issues/7982

Mister-Hope:
Done

PR #1328, opened 2023-05-11
feat(bundler-vite): bump vite to 4.3 and temporarily remove manualChunks

Mister-Hope:


meteorlxy:
Maybe also bump vite version? I reverted it in the previous PR.

PR #1353, opened 2023-06-08
feat(client): add VPLink and RouterLink

Mister-Hope:
## `@vuepress/client`

- new `useResolvedRoute` composables (with redirects support)

two replacement for `<RouterLink>` in `vue-router`:

- new `<VPLink>` component, do not resolve any route during initialization.
- new `<RouterLink>` component, handling redirects and reduce `router.resolve` cost comparing with official `<RouterLink>` and active state with `vp-active`

## `@vuepress/markdown`

- support and use `<VPLink>` by default for links plugin

## `@vuepress/shared`

- new `inferRoutePath` function to avoid possible redirect matching that needs to call `router.resolve` again

## Documentation

All related api docs update

### ( under 6x slowdown mocking withofficial docs)

Original:

Before:

image

After:

image

Adding 1000 pages on official docs:

```ts

  // all pages have been loaded after initialization
  onInitialized: async (app) => {
    for (let i = 1; i < 1000; i++) {
      const homepage = await createPage(app, {
        path: `/${i}`,
        // set frontmatter
        frontmatter: {
          layout: 'Layout',
        },
        // set markdown content
        content: `\
# Welcome to ${app.options.title}

This is page ${i}
`,
      })
      // add it to `app.pages`
      app.pages.push(homepage)
    }
  },
```

Before: 
image

image

After:
image

image


packages/shared/src/utils/inferRoute.ts
```diff
@@ -0,0 +1,18 @@
+// With large amount of routes, calling `router.resolve` can be
+// expensive, so we now try to infer the final path to avoid meeting redirects
+// that needs to resolve the route again
+export const inferRoutePath = (path: string): string => {
```
meteorlxy, line 4: With this util, I think we can remove those redirect routes from pageRoutes.

meteorlxy:
So the new `<RouterLink>` is not actually in use, right?

Is it possible to get rid of both new and old `<RouterLink>`? If so, I think it's better to use `<VPLink>` to replace the `<RouterLink>` totally.

Mister-Hope:
I think we can stop providing the official router-link amd use ours instead. But that coukd be a breaking change.

The router link component should be kept as we do care about active state and the exact href location sometimes.

PR #1357, opened 2023-06-09
fix(plugin-theme-data): fix theme-data sensitive with route object

Mister-Hope:
`useThemeLocaleData()` is sensitive with route object (including route.hash and route.path), so everything based on it will be recalculated when scrolling page and switching pages in same locale.

This issue is related to https://github.com/vuepress/vuepress-next/pull/1253, while that fix is valid for routeLocale itself, but not hacky way `plugin-theme-data` is doing. This PR adds extract check by watching routeLocale ref on `app._context`

This bug has serious performance impact on large sites, please merge it and release new version.


Mister-Hope:
Any update with this?

meteorlxy:
Closed by 9d85d4f5b4fe0bc4bca392b466e168240d156dd3

PR #1359, opened 2023-06-10
chore: update tsconfig

Mister-Hope:
- Node16 targets ES2021
- https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax supported by typescript to force marking types, so we no longer need the eslint rule which is slower.


meteorlxy:
The `verbatimModuleSyntax` does far more than force marking types.

As I remember, it will fail if you write `export default` in source code when output cjs file.

Our project is pure esm so it should be fine to enable it.

PR #1360, opened 2023-06-10
fix(plugin-active-header-links): remove force replace as it updates w…

Mister-Hope:
…hole route object


meteorlxy:
Don't know why we use `force: true` at that time 😅

PR #1361, opened 2023-06-10
fix(client): avoid mismatching between route path and page data (close #1249)

### LINKED ISSUE #1249: [Bug report] Route path is not correct when navigating
### Description

When accessing `route.path` at `onMounted` lifecycle, the value is the previous page, and this is not an expected behavior:

- watcher for route will be triggered unexpected
- a component will initialized with a route path it doesnot belong to

A normal vue app is working as expected: https://stackblitz.com/edit/vitejs-vite-ssyfup?file=src%2Fcomponents%2FB.vue,src%2Fcomponents%2FC.vue,src%2Fcomponents%2FHome.vue&terminal=dev

### Reproduction

https://stackblitz.com/edit/vuepress-ewt9uy?file=docs%2Fguide%2FREADME.md

### Used Package Manager

npm

### System Info

```shell
Not needed
```

Mister-Hope:
Use sync afterEach to update route and pageData in a sync process.

### DIFF
--- a/packages/client/src/router.ts
+++ b/packages/client/src/router.ts
@@ -33,23 +33,23 @@ export const createVueRouter = (): Router => {
     },
   })
 
-  let pendingData: PageData
-  let pendingLocation: string
-
+  // ensure page data and page component have been loaded in beforeResolve hook,
+  // but do not assign page data immediately to avoid mismatching between page data and route path.
+  let pendingPageData: PageData
+  let pendingToPath: string
   router.beforeResolve(async (to, from) => {
     if (to.path !== from.path || from === START_LOCATION) {
-      // ensure page data and page component have been loaded
-      ;[pendingData] = await Promise.all([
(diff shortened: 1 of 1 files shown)


meteorlxy:
@Mister-Hope If I understand correctly, if we update page data in beforeResolve, there might be a short mismatch between route path and page data, because we are using an async callback.

So you move the assignment to afterEach with a sync callback, right?

What if users add another afterEach async callback themselves? Maybe the mismatch would still happen?

Mister-Hope:
I think that hook will be after this one, and don't affect the page data update time.

Mister-Hope:
> What if users add another afterEach async callback themselves? Maybe the mismatch would still happen?

The router does not care afterEach hooks, it just call them directly with `forEach` (regradless they return promise or not).

I tested locally and it does not mismatch. @meteorlxy Maybe apply this fix with vite 4.4 and a new version?

PR #1372, opened 2023-07-04
Add note about VitePress to README.md

KnorpelSenf:

- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Other

### Description

VitePress is so much better already, in so many respects. We recently migrated our highly customised site with ~370 pages and have not regretted it since. Given the slow development speed of VuePress, it feels like the responsible thing to do to make people aware of VitePress, so that they don't end up being stuck here.

I would understand it if you don't want to use VuePress as a vehicle to advertise the competing VitePress project, but IMHO the benefits of VitePress outweigh any concerns regarding fairness—especially because VitePress is more of an upgrade and less or a competitor to VuePress.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,10 @@
 
 ## Status: Beta
 
+> VuePress is now maintained by the community.
+> The core team has moved on to its successor, [VitePress](https://vitepress.dev/).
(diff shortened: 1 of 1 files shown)


meteorlxy:
Thanks, but there is already a tip in the v1 repo.

PR #1418, opened 2023-10-11
chore: update deps

Mister-Hope:


meteorlxy:
Thanks. I'll handle deps bump later.

PR #1433, opened 2023-11-20
feat(client): use visual routes

Mister-Hope:


Mister-Hope:
This PR avoids:

1. heavy operation with `router.resolve`

  `router.resolve` performs `RegExp.exec()` when mapping the whole routes array to find a matched result.

2. duplicated route records with redirects and maps holding data, components, now a single map holds all pages info.


This pr reduce outputsize and improves performance at the same time.

Breaking changes:

- `routeMeta` was renamed to `meta`
- Getting page meta with `router.resolve` is deprecated, people should migrate to `resolve` in `@vuepress/client` to get page meta. (Calling `router.resolve` to get a correct page path is still supported)

packages/client/src/router.ts
```diff
@@ -34,11 +40,22 @@ export const createVueRouter = (): Router => {
...
+      // if no match at this point, then we should provide 404 page
+      const pageInfo = pagesMap.get(pagePath) || pagesMap.get('/404.html')!
+
+      // TODO: Added for backwards compatibility, remove in stable version
+      to.meta = pageInfo.meta
       ;[to.meta._data] = await Promise.all([
-        resolvers.resolvePageData(to.name as string),
-        pagesComponents[to.name as string]?.__asyncLoader(),
+        resolvers.resolvePageData(pageInfo, pagePath),
+        pageInfo.comp?.__asyncLoader(),
```
Mister-Hope, line 58: Notes:

Since we use beforeResolve to align pageData with routePath, we should avoid avoid callindg router.resolve.

This means we should avoid using `router-link` as it internally calls `router.resolve` to provide correct link on anchor tag.

So https://github.com/vuepress/vuepress-next/pull/1353 is still needed.

Though we no longer need RouterLink in that PR, VPLink should still be used by default in markdown and navbar, sidebar links.

Mister-Hope: @meteorlxy I prefer splitting the VPLink PR by  updating the existing one #1353 after you merge this.

packages/shared/src/utils/normalizePath.ts
```diff
@@ -0,0 +1,13 @@
+export const normalizePath = (path: string): string => {
+  const convertedMdPath = path.endsWith('README.md')
+    ? path.substring(0, path.length - 9)
+    : path.endsWith('.md')
+      ? path.substring(0, path.length - 3) + '.html'
+      : path
```
Mister-Hope, line 6: BTW, if we drop markdown link support, we can have higher performance here.

I would upvote to drop support for markdown links, since we already convert every markdown link in markdown files to `<RouterLink>` (or `<VPLink>` with VPLink pr).

At lease we can add an option here?

P.S: What do you think if we add a `router` field in config files with options like :


```
router: {
  cleanUrl: true,
  resolveMarkdownLink: true,
  // ...
}
```

PR #1435, opened 2023-11-20
feat(client): add VPLink

Mister-Hope:
This avoids resolving page data of the destination path with `<router-link>`

Also the built-in `resolve` in `@vuepress/client` has higher performance than `<router-link>`

Note: For performance consideration, `<VPLink />` does not have a built-in active state. You should handle this yourself while using it.

This PR is based on the new router built, and replace #1353


Mister-Hope:
Ci result while finishing official docs page-render stage:

Before: 2.5-3.9s
After: 1.4-2.8s

Local test result with a 1 million words project <https://innenu.com>:

Before: 186.22s
After: 108.11s

PR #1444, opened 2023-12-01
Update vercel deploy docs

sunheyi6:

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

### Description

If you use docs, you will find that the dist folder cannot be found
image

### Screenshots

**Before**

**After**

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -219,6 +219,6 @@ See [Edgio Documentation > Framework Guides > VuePress](https://docs.edg.io/guid
 
    - **FRAMEWORK PRESET:** `Other`
    - **BUILD COMMAND:** `pnpm docs:build`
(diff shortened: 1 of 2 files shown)


meteorlxy:
@sunheyi6  Thanks a lot! However, we just migrated docs into [another repo](https://github.com/vuepress/docs). Would mind create a new PR in that repo? Or we might help to implement this change, but it would be better not to omit your contribution. Thanks!

sunheyi6:
> @sunheyi6 Thanks a lot! However, we just migrated docs into [another repo](https://github.com/vuepress/docs). Would mind create a new PR in that repo? Or we might help to implement this change, but it would be better not to omit your contribution. Thanks!

ok. i will do it.

PR #1445, opened 2023-12-02
fix(theme-default): code group accessibility

nruffing:

- [X] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

### Description

While doing some accessibility testing on [my site](https://datagridvue.com) written in the RC release of Vuepress 2 I came across some accessibility issues with the code group and code group item components.

I have been testing with the latest version of the [axe chrome extension](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd?utm_source=deque.com&utm_medium=referral&utm_campaign=axe_hero).

It reported on the code block item component that `Elements must only use supported ARIA attributes`.  
https://dequeuniversity.com/rules/axe/4.8/aria-allowed-attr?application=AxeChrome

image

The code group item component renders the [`aria-selected`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected) attribute which should only be on elements with the roles `gridcell`, `cell`, `row`, or `tab`. 

Looking through the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#example) at ARIA attributes and tab roles the code group item should actually have the `tabpanel` role and the button to select the tab should have the `tab` role.

The changes in this PR were modeled after the example on MDN. Also, since the roles have requirements of being direct parent/child of each other I removed the wrapping `ul` and `li` elements and only had to slightly alter the CSS to keep the same visual result.

I also added a few color variables for some hard coded colors while I was in there.

### Screenshots

**Before**

image

**After**

image

### Other Accessibility Issues

The only other accessibility I came across was that all of the heading anchors have `aria-hidden` attributes but are still focusable via keyboard navigation.  There was a bit of discussion around this already in the [`markdown-it-anchor` repo](https://github.com/valeriangalliat/markdown-it-anchor/issues/82) and you can address this by just using the `headerLink` render function instead of the `ariaHidden` render function since it wraps the entire header in an anchor.  I applied this change in my [site config directly](https://github.com/nruffing/data-grid-vue/commit/f18a722c804510722448a31b5466868a7c052f11#diff-318da1a121dbfecce693a1fdbc487a7538f06107c09919efaf06706b5f104c00) but I am happy to create a PR to make that the default behavior if that is desired.

image

### DIFF
--- a/ecosystem/theme-default/src/client/components/global/CodeGroup.ts
+++ b/ecosystem/theme-default/src/client/components/global/CodeGroup.ts
@@ -128,38 +128,29 @@ export const CodeGroup = defineComponent({
       return h('div', { class: 'code-group' }, [
         h(
           'div',
-          { class: 'code-group__nav' },
-          h(
-            'ul',
-            { class: 'code-group__ul' },
-            items.map((vnode, i) => {
-              const isActive = i === activeIndex.value
-
-              return h(
-                'li',
-                { class: 'code-group__li' },
-                h(
-                  'button',
-                  {
-                    ref: (element) => {
-                      if (element) {
-                        tabRefs.value[i] = element as HTMLButtonElement
-                      }
-                    },
-                    class: {
-                      'code-group__nav-tab': true,
-                      'code-group__nav-tab-active': isActive,
-                    },
-                    ariaPressed: isActive,
-                    ariaExpanded: isActive,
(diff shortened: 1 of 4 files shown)


nruffing:
No problem. I will create a PR in the new repo.

PR #1447, opened 2023-12-04
perf: implement custom routes

Mister-Hope:
## Description

This pr reduces output size and improves performance at the same time.

Including:

### merging page data and component into 1 chunk

Usually page data chunk is small (<1kb), and with `shouldPreload` `shouldPrefetch` and HTTP1.1, it will affect network. Merging them to 1 chunk almost reduce half of the chunk amount. And do not have negetive influencea as they are always fetch together.

### avoid **heavy operation** with `router.resolve`

- `router.resolve` performs `RegExp.exec()` when mapping the whole routes array to find a matched result.

   To solve this, we introduce:
 
   - virtual routes
   - `hasPage` (repalcement of detecting whether a route exisits)
   - `getPagePaths` (replacement of `routes.getRoutes`)

- To align `route` changes with `pageData`, we exec page component and page data fetching in `beforeResolve` hook, which means every `router.resolve` (so is `<RouterLink>`) actually fetchs that page component and data and parse them.

   To solve this, we introduce:

   - `<VPLink>` component which is "light"
   - `resolve` helper (replacement of `router.resolve` to get path and meta data)

### Removing duplicated route records with redirects and maps holding data, components

Now a single map holds all pages info and pageKey and a redirection map are used at client:

- pagesMap: holds page path and its data, component and meta
- redirectsMap: holds redirects

## Breaking changes

- pageKey is removed at client, so operations like `router.push(pageKey)` `router.resolve(pageKey)` no longer work

- `routeMeta` is renamed to `meta`

  We want to decouple meta concept with router

  `routeMeta` is deprecated but still supported for backwards compatibility, we should remove the support in stable.

- `dataFileChunkName` `dataFilePath` and `dataFilePathRelative` are renamed to `chunkName` `chunkFilePath` and `chunkFilePathRelative`

## New Recommendations

- People should use `resolve` in `@vuepress/client` to get page meta

   Getting page meta with `router.resolve` is deprecated but still supported for backwards compatibility, we should remove the support in stable. (We can support this for sure, but I think that decouple meta with routes could be better)

   Note: Calling `router.resolve` to get a correct page path is still supported as this is a correct usage. (but since it's heavy, maybe we should warn users to use `resolve` as first choice)

- People should use `<VPLink>` as first choice then `<RouterLink>`

  - Links in markdown pages only needs a correct href, ranther then activate status.
  - Links in navbar and sidebar needs a customized active state

  In these cases, they should use `<VPLink>` as the cost of it is 90% less than `<RouterLink>`

## Other notes

To support clean url (e.g.: navigating to `/a/b` and resolve `/a/b.html` to `/a/b`), we improved generation of `htmlRelativePath` and `htmlPath` on Page object

## Benchmark

Ci result while finishing official docs page-render stage:

|Before(main)| After|
|-|-|
| 2.5-3.9s | 1.4-2.8s|

Local test result with a 1 million words project https://innenu.com/:

|Before(main)| After|
|-|-|
| 186.22s |  108.11s|

## RFC

BTW, if we drop markdown link support, we can have higher performance.

I would upvote to drop support for markdown links, since we already convert every markdown link in markdown files to `<VPLink>` with `.html` suffix.

What do you think if we add a router field in config files with options like:

```js
router: {
  cleanUrl: true,
  resolveMarkdownLink: true,
  // ...
}
```

### DIFF
--- a/packages/core/src/types/page.ts
+++ b/packages/core/src/types/page.ts
@@ -112,19 +112,19 @@ export type Page<
     componentFilePathRelative: string
 
     /**
-     * Page file path
+     * Chunk file path
      */
     chunkFilePath: string
 
     /**
-     * Page file path relative to temp directory
+     * Chunk file path relative to temp directory
      */
     chunkFilePathRelative: string
 
     /**
-     * Page file chunk name
+     * Chunk name
      *
-     * Only take effect in webpack
+     * This will only take effect in webpack
      */
     chunkName: string
 


Mister-Hope:
LGTM

packages/client/src/router/resolveRoute.ts
```diff
@@ -0,0 +1,25 @@
...
+  path: string,
+): ResolvedRoute<PageMeta> => {
+  const routePath = resolveRoutePath(path)
+  const pageMapItem = routes.value[routePath] || routes.value['/404.html']
+
+  return {
+    ...(pageMapItem as ResolvedRoute<PageMeta>),
+    path: routePath,
+  }
+}
```
Mister-Hope, line 25: This method is problematic, in a lot of cases, people expect to get `null` when the given path does not belong to any route.

I would suggest a second arguement being added:

```suggestion
export const resolveRoute = <
  PageMeta extends PageMetaDefault = PageMetaDefault,
>(
  path: string,
  fallback = false,
): ResolvedRoute<PageMeta> => {
  const routePath = resolveRoutePath(path)
  const pageMapItem = routes.value[routePath]

  if (!pageMapItem && !fallback) return {
    path: routePath,
    meta: null,
    loader: null,
  };

  return {
    ...(pageMapItem || routes.value['/404.html'] as ResolvedRoute<PageMeta>),
    path: routePath,
  }
}
```

meteorlxy: It's weird to have null meta and loader in a resolved route. I kind of prefer to return `ResolvedRoute<PageMeta>  | null`

Mister-Hope:
LGTM

meteorlxy (approved):

PR #1452, opened 2023-12-09
feat(markdown): update default markdown-it-anchor permalink function (close #1363)

### LINKED ISSUE #1363: [Bug report] The header anchor should stay outside of the heading tag
### Description

The header anchors `#` are placed inside the heading tags (h1, h2, h3...).
This behaviour is not actually correct, since the heading tag should explicitly indicate the nature of the content as a title, and `#` shouldn't be part of it.

Screenshot 2023-06-15 at 00 29 02

Apart from the semantic issue, there is also a more problematic issue: crawlers, some social media crawler may take the title from the h1 heading tag to create a post preview. This means that the preview title of the post will contain the "#" character.

This is an example of a VuePress post preview on LinkedIn

Screenshot 2023-06-15 at 00 34 02

To reproduce I have a live link
https://www.mida.org/posts/reactive-programming-in-financial-markets/

The website above is built with VuePress 1 but the issue persists also on VuePress 2.

Thank you,
Vasile

### Reproduction

https://www.mida.org/posts/reactive-programming-in-financial-markets/

### Used Package Manager

npm

### System Info

```shell
Not relevant
```

nruffing:

- [X] Read the [Contributing Guidelines](https://github.com/vuepress/vuepress-next/blob/main/CONTRIBUTING.md).

- [X] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

### Description

The heading anchors have `aria-hidden` attributes but are still focusable via keyboard navigation.  There was a bit of discussion around this already in the [`markdown-it-anchor` repo](https://github.com/valeriangalliat/markdown-it-anchor/issues/82) and you can address this by just using the `headerLink` render function instead of the `ariaHidden` render function since it wraps the entire header in an anchor.

This fixes #1363 

### Screenshots

We will also need to make some changes to `theme-default`'s base styles to accompany these changes. I will put that PR up shortly and include screenshots there.

**Before**

**After**


meteorlxy:
cc @Mister-Hope any thoughts about this? It should be kind of breaking change for community themes.

Mister-Hope:
I actually do not care about these kind of changes,  This fix the a11y issue, so I am positive towards it.

PR #1456, opened 2023-12-26
feat(markdown): add title support for code blocks (close #1277)

### LINKED ISSUE #1277: Add title support for code blocks

It would be great if people can set a file name for code blocks, e.g.
````md
```js [.vuepress/config.js] {1,3}
const a = 1;

console.log(“hello world”)
```
````

### Suggested solution

Above

### Alternative

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by nruffing**

- [X] Read the [Contributing Guidelines](https://github.com/vuepress/vuepress-next/blob/main/CONTRIBUTING.md).

- [ ] Bug fix
- [X] New feature
- [ ] Documentation update
- [ ] Other

### Description

Add support for being able to set a title for code blocks (e.g. a filename).

In case the `data-ext` attribute was being used for something else in a theme I added a new `data-title` attribute for the title and its value falls back to be the same as `data-ext` if an explicit title is not specified.

I opted to use enclose the title in `<` and `>` instead of square brackets so it would also easily work with the import code plugin which is already using square brackets.

#### Example
````md
```ts <config/foo.ts> {1,2}:no-line-numbers
const foo = 'foo'

function bar () {
  return 1024
}
```
````

Tests have also been updated.

If we like this approach I can create a PR in `theme-default` to change the css to pull from the `data-title` attribute and update the docs.

closes #1277 

### Screenshots

**Before**

**After**

### CODE REVIEW

**Review by Mister-Hope — APPROVED** [review]

**Comment by meteorlxy** [comment]
Any reference of the `[]` and `<>` of other markdown extensions? I think it would be better to use a commonly used mark so that users could migrate from other framework to vuepress with ease

**Comment by nruffing** [comment]
Here are some examples I found with some quick research. Based on this I could be on board with a key/value syntax like `title="config/foo.ts"`.

## [fenceparser](https://github.com/frencojobs/fenceparser)

````md
```ts twoslash {1-3, 5} title="Hello, World"
```
````

## [MyST](https://mystmd.org/guide/code#showing-a-filename)

````md
```{code} yaml
:filename: myst.yml
project:
  title: Showing Filenames in code-blocks
```
````

## [Docusaurus](https://docusaurus.io/docs/markdown-features/code-blocks)

````md
```jsx title="/src/components/HelloCodeTitle.js"
function HelloCodeTitle(props) {
  return <h1>Hello, {props.name}</h1>;
}
```
````

## [rehype-code-titles](https://unifiedjs.com/explore/package/rehype-code-titles/#input)

````md
```typescript:lib/mdx.ts
```
````

meteorlxy:
Agree with the key-value implementation. It's also a more extendable syntax

nruffing:
Sounds good. I will make those changes.

PR #1460, opened 2023-12-29
feat(vuepress): add client type export

Mister-Hope:


packages/vuepress/package.json
```diff
@@ -25,6 +25,7 @@
     "./cli": "./dist/cli.js",
     "./client": "./dist/client.js",
     "./client-app": "./dist/client-app.js",
+    "./client-types": "./dist/client-types.js",
```
meteorlxy, line 28: Where's the client-types file? 😅

meteorlxy (changes requested):

packages/vuepress/package.json
Mister-Hope: oh the dts file, forget to add that

meteorlxy (approved):

PR #1475, opened 2024-01-24
fix(bundler-webpack): add conditionNames to import correct file

Mister-Hope:
See https://webpack.js.org/configuration/resolve/#resolveconditionnames

This solves a lot of issues.

I previously use alias with conditional checks (app.env.isDev, app.env.isBuild and bundler names) to  ensure the correct file is used, but this can obviously fixed upstream here.

For example, when importing `@vue/repl` with webpack, `@vue/repl` import `vue/component-sfc` in it. And the latter use a condition export like this:

```json
  "exports": {
    ".": {
      "types": "./dist/compiler-sfc.d.ts",
      "node": "./dist/compiler-sfc.cjs.js",
      "module": "./dist/compiler-sfc.esm-browser.js",
      "import": "./dist/compiler-sfc.esm-browser.js",
      "require": "./dist/compiler-sfc.cjs.js"
    },
  },
```

Then webpack should pack `./dist/compiler-sfc.esm-browser.js` for it, and use `"./dist/compiler-sfc.cjs.js"` for SSR.

Another case can be `bcrypt-ts` 

```json
  "exports": {
    ".": {
      "types": "./dist/node.d.mts",
      "browser": "./dist/browser.mjs",
      "node": "./dist/node.mjs",
      "import": "./dist/node.mjs",
      "require": "./dist/node.cjs",
      "default": "./dist/node.cjs"
    },
  },
```

The crypto module is different under node ('node:crypto') and browser (window.Crypto)

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -17,6 +17,9 @@ module.exports = {
     {
       files: ['**/e2e/**/*.cy.ts', '**/e2e/cypress/**/*.ts'],
       extends: 'plugin:cypress/recommended',
+      rules: {
+        '@typescript-eslint/no-namespace': 'off',
+      },
     },
     {
       files: ['**/tests/**/*.ts', 'tsup.config.ts'],
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -64,6 +64,7 @@ jobs:
         env:
           E2E_BASE: ${{ matrix.base }}
           E2E_BUNDLER: ${{ matrix.bundler }}
+          E2E_COMMAND: dev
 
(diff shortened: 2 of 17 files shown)


meteorlxy:
@Mister-Hope Any minimal reproduction for this issue? I could not reproduce it when trying to add e2e cases

Mister-Hope:
https://github.com/vuepress-theme-hope/vuepress-theme-hope/actions/runs/7513512110/job/20455377640

meteorlxy:
@Mister-Hope Could help to create a valid reproduce in `imports/conditional-exports` e2e test?

PR #1521, opened 2024-03-25
feat(client): support relative path (close #1492)

### LINKED ISSUE #1492: [Bug report] resolveRoute does not support relative path
### Description

`resolveRoute` should support relative path, just like `router.resolve()`

E.g.:

```ts
// at /page/a.html
resolve('b.html') // should be /page/b.html
resolve('../') // should be /
```

## Why this is needed

prev and next links are based on current route path, as people just set relative links like `./a.md` and `../`

Mister-Hope:
This adds:

- logic changes on `isLinkHttp`, now link starts with `//` is considered as http link
- logic changes on `isLinkExternal`, now link with protocols will all be treated as external
- relative link support for `normalizeRoutePath` `resolveRoutePath` and `resolveRoute`
- A new component `<AutoLink>` which can be used to render links in plugins and themes


Mister-Hope:
@meteorlxy Could you review this? The downstream needs this.

packages/client/src/components/RouteLink.ts
```diff
@@ -23,72 +23,75 @@ const guardEvent = (event: MouseEvent): boolean | void => {
...
+  }>,
+
+  setup(props, { slots }) {
+    const router = useRouter()
+    const route = useRoute()
+
+    const path = computed(() => {
+      if (
+        // as is
+        ['#', '?', 'http://', 'https://', '//'].some((prefix) =>
```
meteorlxy, line 72: Maybe extract to a util function? Or at least extract the array to a constant.

Also need test cases for links with protocols

packages/client/src/router/resolveRoutePath.ts
```diff
@@ -1,12 +1,22 @@
...
  */
-export const resolveRoutePath = (path: string): string => {
+export const resolveRoutePath = (path: string, current?: string): string => {
+  let routePath = path
+
+  if (!path.startsWith('/') && current) {
+    const loc = current.slice(0, current.lastIndexOf('/'))
+
+    routePath = new URL(`${loc}/${path}`, FAKE_HOST).pathname
+  }
```
meteorlxy, line 16: Better to extract to a standalone file like `resolveRelativePath`. Maybe put it to shared pkg and add unit tests for it.

packages/shared/tests/normalizeRoutePath.spec.ts
```diff
@@ -42,28 +42,138 @@ const testCases = [
...
-  ['foo.md?bar=baz', 'foo.html?bar=baz'],
-  ['foo.html?bar=baz', 'foo.html?bar=baz'],
-  ['foo?bar=baz#qux', 'foo.html?bar=baz#qux'],
-  ['foo.md?bar=baz#qux', 'foo.html?bar=baz#qux'],
-  ['foo.html?bar=baz#qux', 'foo.html?bar=baz#qux'],
-  ['#bar', '#bar'],
-  ['?bar=baz', '?bar=baz'],
-  ['?bar=baz#qux', '?bar=baz#qux'],
+  // only hash and query
+  ['', ''],
```
meteorlxy, line 46: Cases for hash and query are missed?

Mister-Hope: they are undet the.below.test

PR #1525, opened 2024-04-01
fix(client): export ResolveRoute

Mister-Hope:


Mister-Hope:
This type is returned by `resolveRoute`, so we'd better export it.

meteorlxy (approved):

PR #1532, opened 2024-04-07
perf(client): avoid unnecessary parsing

Mister-Hope:


meteorlxy:
Any reference? `JSON.parse` should have better performance than loading plain object

PR #1537, opened 2024-04-11
fix(markdown): fix query parsing and absolute link fallback in linksPlugin (close #1536)

### LINKED ISSUE #1536: [Bug report] linkPlugin can not handle query
linkPlugin can not handle internal link query

Mister-Hope:
Closes #1536

- Now queries are preserved instead of being removed
- Absolute link is `null` when a link is relative and there is no markdown file info

This PR is based on #1521


packages/client/src/components/AutoLink.ts
```diff
@@ -0,0 +1,227 @@
...
+import { computed, defineComponent, h } from 'vue'
+import { useRoute } from 'vue-router'
+import { useSiteData } from '../composables/index.js'
+import { RouteLink } from './RouteLink.js'
+
+export interface AutoLinkConfig {
+  /**
+   * Text of item
+   *
+   * 项目文字
```
meteorlxy, line 12: remove zh comments

meteorlxy:
Need to update test cases

meteorlxy (changes requested):
Need to update test cases

Mister-Hope:
What test cases is not update? This PR is not introducing any new functions, while I already updated existing ones.

meteorlxy (approved):

PR #1544, opened 2024-04-16
feat(shared): support relative links in normalizeRoutePath

Mister-Hope:


packages/shared/src/utils/normalizeRoutePath.ts
```diff
@@ -1,15 +1,11 @@
-/**
- * Normalize the given path to the final route path
- */
-export const normalizeRoutePath = (path: string): string => {
-  // split pathname and query/hash
-  const [pathname, ...queryAndHash] = path.split(/(\?|#)/)
+const FAKE_HOST = 'http://.'
 
+export const inferRoutePath = (path: string): string => {
```
meteorlxy, line 3: As it's exported, we'd better split it into a standalone file

packages/shared/tests/normalizeRoutePath.spec.ts
```diff
@@ -15,55 +15,168 @@ const testCases = [
...
-  ['foo.md?bar=baz', 'foo.html?bar=baz'],
-  ['foo.html?bar=baz', 'foo.html?bar=baz'],
-  ['foo?bar=baz#qux', 'foo.html?bar=baz#qux'],
-  ['foo.md?bar=baz#qux', 'foo.html?bar=baz#qux'],
-  ['foo.html?bar=baz#qux', 'foo.html?bar=baz#qux'],
-  ['#bar', '#bar'],
-  ['?bar=baz', '?bar=baz'],
-  ['?bar=baz#qux', '?bar=baz#qux'],
+  // relative non index with current
+  ['foo', '/foo.html', '/'],
```
meteorlxy, line 49: Better to use `[[input1, input2], output]` format, instead of putting the second param  as the third item of the tuple

meteorlxy (dismissed):

packages/shared/tests/normalizeRoutePath.spec.ts
```diff
@@ -2,109 +2,232 @@ import { describe, expect, it } from 'vitest'
...
 
   // unexpected corner cases
-  ['.md', '.html'],
-  ['foo/.md', 'foo/.html'],
-  ['/.md', '/.html'],
-  ['/foo/.md', '/foo/.html'],
+  [['.md'], '.html'],
+  [['foo/.md'], 'foo/.html'],
+  [['/.md'], '/.html'],
+  [['/foo/.md'], '/foo/.html'],
```
meteorlxy, line 185: Still need cases with current for the last 5 lines

Mister-Hope: I actually don't think this is needed,  These are corner cases which we are not expecting users to actually set them in production

meteorlxy: They are unexpected, but still should be covered by our unit tests

meteorlxy: For example, if a method is going to throw an error with unexpected input, we usually test it with `expect().toThrow()`. For this case, we don't throw errors, but we still need to be aware of those unexpected input.

Mister-Hope:
test added

meteorlxy (approved):

PR #1545, opened 2024-04-16
feat(client): support relative link in RouteLink

Mister-Hope:


Mister-Hope:
cc @meteorlxy

Mister-Hope:
Why is the e2e test failing again... 😒😒

meteorlxy (approved):

PR #1546, opened 2024-04-17
feat(client): add AutoLink component

Mister-Hope:


packages/client/src/components/AutoLink.ts
```diff
@@ -0,0 +1,227 @@
...
+import { computed, defineComponent, h } from 'vue'
+import { useRoute } from 'vue-router'
+import { useSiteData } from '../composables/index.js'
+import { RouteLink } from './RouteLink.js'
+
+export interface AutoLinkConfig {
+  /**
+   * Text of item
+   *
+   * 项目文字
```
meteorlxy, line 12: remove zh comments

Mister-Hope:
See here https://github.com/vuejs/router/blob/8fbd9c427447a8e4a6ff486ae94224d040fb7b00/packages/router/src/RouterView.ts#L34

I strongly recommend preserving the prop interface.

meteorlxy (approved):

PR #1547, opened 2024-04-17
feat(core): support permalink: null in frontmatter to disable permalink

Mister-Hope:
Now users can set permalink: null to disable global `permalinkPattern` affect on certain pages.

This is more natural then setting permalink or permalinkPattern to page path in frontmatter.


meteorlxy (changes requested):
Need to update unit tests

meteorlxy (approved):

PR #1554, opened 2024-05-16
feat(client): improve AutoLinkProps

Mister-Hope:


Mister-Hope:
Downstream env will often provide complicated config object, e.g.:

```json
{
  "text": "ABC",
  "link": "/ABC.html",
  "icon": "abc"
  "children": [
  // ...
  ]
]
```

For the current prop, developers and users must declare every props, otherwise the default `inheritAttrs` behavior will make extra configuraion be element props, and compilated ones will become `Object object`, further more, when setting element preserved keys (e.g.: children), warnings are disaplyed by Vue telling those keys only have getter.

meteorlxy (approved):

PR #1562, opened 2024-05-19
fix: fix route resolving error with hash and queries (close #1561)

### LINKED ISSUE #1561: [Bug report] When the page path includes `hash`/ `query`, the page will load a 404 error.
### Description

After upgrading from `rc.9` to `rc.11`, there was a change in `beforeResolve` where the path comparison switched from `to.path` to `to.fullpath`. 

However, `to.fullpath` includes the `hash` portion, which causes issues when the page is first loaded or when navigating to a page with a `hash` in the path, resulting in the inability to match the correct route and leading to a `404` error.

image

### Reproduction

-

### Used Package Manager

pnpm

### System Info

```shell
vuepress: 2.0.0-rc.11 => 2.0.0-rc.11
```

pengzhanbo:

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

### Description

fix: #1561


Mister-Hope (dismissed):

packages/client/src/router/resolveRoute.ts
```diff
@@ -16,7 +17,8 @@ export const resolveRoute = <T extends RouteMeta = RouteMeta>(
   currentPath?: string,
 ): ResolvedRoute<T> => {
   const routePath = resolveRoutePath(path, currentPath)
-  const route = routes.value[routePath] ?? {
+  const pathname = resolveRoutePathname(routePath)
+  const route = routes.value[pathname] ?? {
```
meteorlxy, line 21: Need e2e tests to track this fix

pengzhanbo: Done.

digital-codes:
changing  fullPath tp path in client/dist/app.js like so fixes the issue with query strings 
...
router.beforeResolve(async (to, from) => {
    if (to.path !== from.path || from === START_LOCATION) {
      const route = resolveRoute(to.path);
...

Not sure what's the exact difference and why it worked up to rc8 (at least) and broke with rc11
Also not sure about side effects ...

you have 
const fullPath = to.fullPath.split('#')[0]

Will fix hash but not query parms, right?

pengzhanbo:
> changing fullPath to path in client/dist/app.js like so fixes the issue with query strings ... router.beforeResolve(async (to, from) => { if (to.path !== from.path || from === START_LOCATION) { const route = resolveRoute(to.path); ...
> 
> Not sure what's the exact difference and why it worked up to rc8 (at least) and broke with rc11 Also not sure about side effects ...
> 
> you have const fullPath = to.fullPath.split('#')[0]
> 
> Will fix hash but not query parms, right?

The bug was caused by the fact that vuepress internally needs to resolve the route to obtain the route instance through `pathname`, but `route.fullPath` contains `pathname`, `query`, `hash`, and other content, leading to parsing errors and failure to match the route instance. 

You can refer to the changes made in this PR submission for more information.

Mister-Hope:
I updated the pr to improve the original fix performance.

Also now API are unified to avoid misleading.

- All parameters which is expected with no hash and queries are renamed as `pathname`
- The above function are named with "Route Path`

- Any parameters which may have hash and queries are called `path`
- The above functions are named with "Route Full Path"

Mister-Hope (approved):

packages/shared/tests/normalizeRoutePath.spec.ts
```diff
@@ -204,42 +204,3 @@ describe('should normalize clean paths correctly', () => {
     }),
   )
 })
-
-describe('should normalize paths with query correctly', () => {
```
meteorlxy, line 208: So we are not keeping hash & queries anymore in `normalizeRoutePath`. Any potential breakings?

packages/shared/tests/routes/normalizeRoutePath.spec.ts
```diff
@@ -204,42 +204,3 @@ describe('should normalize clean paths correctly', () => {
     }),
   )
 })
-
-describe('should normalize paths with query correctly', () => {
```
meteorlxy, line 208: So we are not keeping hash & queries anymore in `normalizeRoutePath`. Any potential breakings?

Also, per [this comment](https://github.com/vuepress/core/pull/1562#issuecomment-2119662222):

> All parameters which is expected with no hash and queries are renamed as pathname

Should this method be renamed to `normalizeRoutePathname`?

Mister-Hope: Nope,  routePath is for route.path, and routeFullPath is for route.fullPath.  The former does not have any hash or queries on it.

For the second one, the answer might be yes,  But before the rc9,  That function does not support harsh or queries,  So it is yes to say that we introduce a potential breaking changes in this pr, but it is likely not to break any functions downstream.

Mister-Hope: I understand that for url definition,  This world should be named as pathname,  But we are all converting them to route  related things and route path and route full path are the things that they should be compared or set.

So I decided to name them as the things I described earlier

e2e/tests/router/navigate-by-link.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { expect, test } from '@playwright/test'
+import { BASE } from '../../utils/env'
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('router/navigate-by-link.html')
+})
+
+test('TODO', async ({ page }) => {
+  // TODO
+  await expect(page).toHaveURL(`${BASE}router/navigate-by-link.html`)
```
meteorlxy, line 10: Made some updates. It would be helpful if you can help to complete the test cases. If you don't have time, I'll add it tmr.

Mister-Hope: Done

meteorlxy (approved):

PR #1563, opened 2024-05-20
fix(core): fix page redirects comparison

Mister-Hope:


Mister-Hope:
The previous code double encode the `pathInfer` by mistake while comparing.

packages/core/src/app/prepare/prepareRoutes.ts
```diff
@@ -27,20 +27,15 @@ const resolvePageRedirects = ({ path, pathInferred }: Page): string[] => {
...
-
   // redirect from inferred path, notice that the inferred path is not uri-encoded
   if (pathInferred !== null) {
-    addRedirect(encodeURI(pathInferred))
+    const normalizedPathInferred = normalizeRoutePath(pathInferred)
+    const encodedPathInferred = encodeURI(normalizedPathInferred)
+
+    // add redirect to the set when the redirect could not be normalized & encoded to the page path
+    if (normalizedPathInferred !== path && encodedPathInferred !== path) {
+      redirectsSet.add(encodedPathInferred)
```
meteorlxy, line 37: Maybe we don't need the set anymore? After several iteration, there's only one possible redirect here?

Mister-Hope: I think we should still compare the boss because users might set a `/中文.html` manually.  And this extra compare would not take much time.

meteorlxy (approved):

PR #1564, opened 2024-05-20
feat(core): allow non-default-export client config file

Mister-Hope:
do not require `export default {}` with client config file

useful for:

- users that only want to add this file for plugin customization purpose
- plugins which only need to add some styles or library imports on the client side with this


meteorlxy (commented):
1. Better to check default export in node side instead of filtering it in client code.
2. Better to add e2e tests for the changes.

Mister-Hope:
I'm not pretty sure about the first one,  So you mean importing this file on the note side and see if the default export is available? 

So what's the motivation of this?  We can't just filter those files which does not have sport because we shall import them anyway to have their library and style file imported.  Also, if the fail is containing some style files import like css and scss,  An area is expected at node side because we are not doing them with bundlers. 

And I think there is no actual performance decrease  for using a namespace import,  Because only the default export is expected on this file.

meteorlxy:
I mean to check the file content in node side or something. As I remember, v1 did similar thing by checking `export default` keyword.

Mister-Hope:
I think  checking without ast may introduce fail positive, so I prefer the original one

Mister-Hope:
shit, why is windows failing (ー_ー)!!

e2e/docs/.vuepress/plugins/clientConfig.js
```diff
@@ -0,0 +1 @@
+import './test.css'
```
meteorlxy, line 1: Should check if the styles are applied correctly to the target elements

meteorlxy (approved):

PR #1565, opened 2024-05-21
chore(bundler-webpack): use webpack internal types

Mister-Hope:


packages/bundler-webpack/src/build/ssr/createClientPlugin.ts
```diff
@@ -51,22 +52,25 @@ export const createClientPlugin = (
 
           // get asset modules
           const assetModules = modules.filter(
-            (m): m is FnModules & Required<Pick<FnModules, 'assets'>> =>
-              !!(m.assets && m.assets.length),
+            (
+              module,
```
meteorlxy, line 56: We'd better avoid using`module` keyword as var name. That's why we use `m` previously

meteorlxy (approved):

PR #1566, opened 2024-05-21
feat(bundler-webpack): replace webpack-chain with webpack-5-chain (close #1503)

### LINKED ISSUE #1503: [Bug report] Replace webpack-chain with webpack-5-chain
### Description

Technically speaking, webpack-chain never adds support for webpack v5, and the latest version was published 4 years ago

The maintainer already deprecates the package, so we'd better drop it.

See https://www.npmjs.com/package/webpack-chain

Mister-Hope:


packages/bundler-webpack/src/build/createClientConfig.ts
```diff
@@ -88,13 +93,11 @@ export const createClientConfig = async (
...
-  config.optimization.set('minimizer', [
-    // keep the default minimizer
-    '...',
-    // add css minimizer
-    new CssMinimizerPlugin({
-      minify: CssMinimizerPlugin.lightningCssMinify,
-    }),
+  config.optimization.minimizer('css').use(CssMinimizerPlugin, [
+    {
+      minify:
```
meteorlxy, line 98: This would remove the default minimizer.

meteorlxy (approved):

PR #1588, opened 2024-07-13
feat(bundler-webpack): add api field in sass-loader option type

Mister-Hope:


meteorlxy:
Closing as `sass-loader` already prefers `sass-embedded` over `sass` since v15.0.0

See: https://github.com/webpack-contrib/sass-loader/pull/1211

meteorlxy:
Oops I misunderstood the use case. This PR is enabling `modern-compiler` by default. It might be kind of a breaking change, and we might wait for vite support before merging this.

PR #1606, opened 2024-09-10
refactor: remove temp page files and load page component via bundler

meteorlxy:

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other
- [x] Refactor

### Description

Changes:
  - refactor: remove temp page files and load page component via bundler
  - refactor(markdown): remove relativePathPrefix option from assetsPlugin
  - feat: support importing module via relative paths in page files
  - feat: support importing markdown as vue component

### Screenshots

N/A


Mister-Hope (commented):
I think there should be other improvements in this PR:

1. Since we already rebuild markdown handling, it's time to support more flexible page generation.

Generally, .vue and component with .js .ts could all be pages, so we can  make vuepress support them out of box, while remaining the default pagesPattern to be only markdown.

2. The ecosystem and downstream plugins need a new option, I would design it as `onPageUpdate`

Technically speaking, a plugin can get all pages in `onInitialized` lifecycle, but when a new page is added, changed or removed in devServer, a plugin cannot subscribe, we are watching the temp files before this PR, but this workaround no longer works now.

So to let the plugin work well, the `onPageUpdate` should be:

```ts
onPageUpdated(app: App, type: 'add' | 'change' | 'remove', oldPage: Page | null, newPage: Page | null)
```

Or we can split it into 3 subscriber:

```ts
onPageAdded(app: App, newPage: Page | null)
onPageChanged(app: App, oldPage: Page, newPage: Page)
onPageRemoved(app: App, oldPage: Page)
```

PR #1613, opened 2024-10-09
feat(core)!: remove headers field from page data

Mister-Hope:
This reduces page data size by removing headers info from page data, downstream themes can infer headers via dom structure with `@vuepress/helper` and they can manually write it to page data if necessary.


meteorlxy:
Then the markdown.headers should default to `false`?

meteorlxy:
Could headers be server-side rendered after the change?

Mister-Hope:
I think the header information might still be required on the page object, but we should no longer pass it to client data by default.

meteorlxy:
I see. What about SSR? Seems that helper pkg is using DOM api to get headers.

Mister-Hope:
A tic is not needed for ssr I think.

Mister-Hope:
I think this can be merged

meteorlxy (approved):

PR #1636, opened 2025-02-25
refactor(client): use set to store content updated callbacks

meteorlxy:

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Other

### Description

Replace array with set to avoid duplicated callbacks

### Screenshots

N/A

### DIFF
--- a/e2e/docs/.vuepress/client.ts
+++ b/e2e/docs/.vuepress/client.ts
@@ -1,5 +1,6 @@
 import { defineClientConfig } from 'vuepress/client'
 import ComponentForMarkdownGlobal from './components/ComponentForMarkdownGlobal.vue'
+import OnContentUpdated from './components/OnContentUpdated.vue'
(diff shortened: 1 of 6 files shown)


Mister-Hope (approved):

Mister-Hope:
I understand this could be an improvement, but are there real cases where a same function could be called multiply by this API?

This is not working:

```ts
const something = () => {
  // ...
}

onContentUpdate(() => {
  something()
})

onContentUpdate(() => {
  something()
  anotherThing()
})
```

meteorlxy:
@Mister-Hope Your example code are not duplicated callbacks. The actual callback is the anonymous arrow function.

The original implementation used `filter` in the `onUnmounted` hook, will which also delete all duplicated callbacks. So I think using set to dedupe callbacks should be as expected for current stage.

PR #1642, opened 2025-04-09
fix(bundler-webpack): allow UPPERCASE file extension

Mister-Hope:


Mister-Hope:
Why this is needed: There are cases where the file extension is UPPERCASE, e.g.: iOS photo app exports `.PNG` by default.

meteorlxy (approved):

PR #1648, opened 2025-05-02
fix: make alias can be overridden properly

Mister-Hope:

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

### Description

Vite/Webpack current has issues handling alias, alias resolution may differ with how they defined.

Users may fail to override some subpaths if a parent folder alias was defined by plugins or themes.

This PR solves the problem by sorting alias key with its length first to ensure longer alias being replaced first.

Also, though webpack support alias as `string[] | false`, Vite can not support these things, so to ensure alias is working in both bundlers, the type definition is restricted to `Record<string, string>`. Users are still allowed to add advanced definition with bundler options.


packages/bundler-vite/src/plugins/vuepressConfigPlugin.ts
```diff
@@ -25,16 +25,16 @@ const resolveAlias = async ({
...
-    })
+    Object.assign(alias, aliasObject)
   })
 
   return [
-    ...Object.keys(alias).map((item) => ({
-      find: item,
-      replacement: alias[item],
-    })),
+    ...Object.keys(alias)
```
meteorlxy, line 32: Add comments to highlight that we need to ensure the order of alias keys.

packages/bundler-webpack/src/config/handleResolve.ts
```diff
@@ -14,17 +14,26 @@ export const handleResolve = async ({
...
-    '.mjs': ['.mjs', '.mts'],
+  // plugin hook: alias
+  const aliasResult = await app.pluginApi.hooks.alias.process(app, isServer)
+
+  aliasResult.forEach((aliasObject) => {
+    Object.assign(alias, aliasObject)
   })
 
+  // set aliases
+  config.resolve.alias.merge(
```
meteorlxy, line 31: Same here

meteorlxy (approved):

PR #1690, opened 2026-03-19
feat(core): add `.git` to default pagePatterns

Mister-Hope:


packages/core/src/app/resolveAppOptions.ts
```diff
@@ -43,7 +43,7 @@ export const resolveAppOptions = ({
   bundler,
   debug = false,
   markdown = {},
-  pagePatterns = ['**/*.md', '!.vuepress', '!node_modules'],
+  pagePatterns = ['**/*.md', '!.vuepress', '!node_modules', '!.git'],
```
meteorlxy, line 46: Cuz `pagePatterns` allows user customization, it might be verbose to always include '!.git'. And that's why it would be a breaking change.

Maybe we could maintain a constant for default ignore pattern. Paths like `.git` should always be ignored (and maybe `node_modules`).

Any thought? @Mister-Hope  @pengzhanbo

Mister-Hope: I can hardcode the two in watchPagesFiles.ts at cli package with a IGNORE_FOLDERS constant for sure.

I agree users may never want to include .git and .node_modules folders.

But you need to merge #1689 first otherwise conflicts will be introduced.

PR #1698, opened 2026-05-08
ci: bump actions and drop node 20

Mister-Hope:


Mister-Hope:
This is now blocked by https://github.com/pnpm/action-setup/issues/150.

A workaround is adding top-level `packageManager` field back.

meteorlxy (approved):

PR #1706, opened 2026-05-18
fix(core): update default value for absolutePathPrependBase

Mister-Hope:


Mister-Hope:
This should be better handled in core, developers like me could forget this.

Users wound not dig into our docs when using Webpack, so this will sounds like a bug to them.

meteorlxy (approved):

PR #1707, opened 2026-05-18
test(e2e): update tests for router and route-link

Mister-Hope:
Current e2e tests are a bit confusing, as everything behind `#` will be treated as hash, so there won't be any "hash + query" urls.

Now, tests are correctly described as complex hash with updated tests

### DIFF
--- a/e2e/docs/router/navigate-by-router.md
+++ b/e2e/docs/router/navigate-by-router.md
@@ -4,7 +4,7 @@
 <button id="home-with-query" @click="goHomeWithQuery">Home</button>
 <button id="home-with-query-and-hash" @click="goHomeWithQueryAndHash">Home</button>
 <button id="not-found-with-hash" @click="go404WithHash">404</button>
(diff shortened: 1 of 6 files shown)


e2e/docs/router/navigate-by-router.md
```diff
@@ -31,7 +31,7 @@ const go404WithHash = () => {
   router.push('/404.html#_404');
 }
 
-const go404WithHashAndQuery = () => {
-  router.push('/404.html#_404?notFound=true');
+const go404WithQueryAndHash = () => {
```
meteorlxy: The name should be go404WithComplexHash then?

Mister-Hope: Yes, I will fix that

Mister-Hope: Done

meteorlxy (approved):

PR #1708, opened 2026-05-19
feat(bundler-helper): add bundler-helper package

Mister-Hope:
This PR migrated bundler related functions from `@vuepress/helper` to `@vuepress/bundler-helper` (a new package in core).

We are going to have 3 bundlers now, so such helper is quite useful.

Also it's better to align the package with core.

Plus, the package solves sevel problems:

1. Providing config and let them being merged to final config automatically. For Webpack this might be easy, but for vite this is not easy, as `mergeViteConfig` function is not available (could be a PR later).
2. It's complex for developers to handle bundler config in plugins and themes:

  Code from official docs ([Ref](https://vuejs.press/reference/plugin-api.html#extendsbundleroptions)): 
  
```ts
export default {
  extendsBundlerOptions: (bundlerOptions, app) => {
    // extends options of @vuepress/bundler-vite
    if (app.options.bundler.name === '@vuepress/bundler-vite') {
      bundlerOptions.vuePluginOptions ??= {}
      bundlerOptions.vuePluginOptions.template ??= {}
      bundlerOptions.vuePluginOptions.template.compilerOptions ??= {}
      const isCustomElement =
        bundlerOptions.vuePluginOptions.template.compilerOptions.isCustomElement
      bundlerOptions.vuePluginOptions.template.compilerOptions.isCustomElement =
        (tag) => {
          if (isCustomElement?.(tag)) return true
          if (tag === 'my-custom-element') return true
        }
    }

    // extends options of @vuepress/bundler-webpack
    if (app.options.bundler.name === '@vuepress/bundler-webpack') {
      bundlerOptions.vue ??= {}
      bundlerOptions.vue.compilerOptions ??= {}
      const isCustomElement = bundlerOptions.vue.compilerOptions.isCustomElement
      bundlerOptions.vue.compilerOptions.isCustomElement = (tag) => {
        if (isCustomElement?.(tag)) return true
        if (tag === 'my-custom-element') return true
      }
    }
  },
}
```

Now:

  
```ts
import { addCustomElement } from 'vuepress/bundler-helper'

export default {
  extendsBundlerOptions: (bundlerOptions, app) => {
    addCustomElement(bundlerOptions, app, 'my-custom-element')
  }
}
```

### DIFF
--- a/packages/bundler-helper/README.md
+++ b/packages/bundler-helper/README.md
@@ -1,0 +1,12 @@
+# @vuepress/bundlerutils
+
+[![npm](https://badgen.net/npm/v/@vuepress/bundlerutils/next)](https://www.npmjs.com/package/@vuepress/bundlerutils)
+[![license](https://badgen.net/github/license/vuepress/core)](https://github.com/vuepress/core/blob/main/LICENSE)
+
+## Documentation
+
+https://vuepress.vuejs.org
+
+## License
+
+[MIT](https://github.com/vuepress/core/blob/main/LICENSE)
--- a/packages/bundler-helper/README.md
+++ b/packages/bundler-helper/README.md
@@ -1,6 +1,6 @@
-# @vuepress/bundlerutils
+# @vuepress/bundler-utils
 
-[![npm](https://badgen.net/npm/v/@vuepress/bundlerutils/next)](https://www.npmjs.com/package/@vuepress/bundlerutils)
+[![npm](https://badgen.net/npm/v/@vuepress/bundler-utils/next)](https://www.npmjs.com/package/@vuepress/bundler-utils)
 [![license](https://badgen.net/github/license/vuepress/core)](https://github.com/vuepress/core/blob/main/LICENSE)
 
 ## Documentation
--- a/packages/bundler-helper/package.json
+++ b/packages/bundler-helper/package.json
@@ -1,0 +1,66 @@
+{
+  "name": "@vuepress/bundler-helper",
+  "version": "2.0.0-rc.30",
+  "description": "Helper package of VuePress bundlers",
+  "keywords": [
+    "bundler",
+    "utils",
+    "vuepress"
+  ],
+  "homepage": "https://github.com/vuepress",
+  "bugs": {
+    "url": "https://github.com/vuepress/core/issues"
+  },
+  "license": "MIT",
+  "author": "Mister-Hope",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/vuepress/core.git"
+  },
+  "files": [
+    "dist"
+  ],
+  "type": "module",
+  "main": "./dist/index.js",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": "./dist/index.js",
+    "./package.json": "./package.json"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "build": "tsdown",
+    "clean": "rimraf dist"
+  },
+  "dependencies": {
+    "@vuepress/core": "workspace:*",
+    "@vuepress/shared": "workspace:*",
+    "@vuepress/utils": "workspace:*"
+  },
+  "devDependencies": {
+    "@types/connect": "^3.4.38"
+  },
+  "peerDependencies": {
+    "@vuepress/bundler-vite": "workspace:*",
+    "@vuepress/bundler-webpack": "workspace:*"
+  },
+  "peerDependenciesMeta": {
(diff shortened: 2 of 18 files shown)


meteorlxy:
- The name pattern `bundler-[bundler-name]` is for bundlers. That's why we name another package as `bundlerutils`. (Similarly, rollup has [pluginutils](https://github.com/rollup/plugins/tree/master/packages/pluginutils))

- Any necessity to put this package into core instead of ecosystem?

Mister-Hope:
1. Bundler updates are made in core, and this aligns support for bundler helper package in each release. We should consider that bundler configs changed in time, so moving this to core ensures there is no support mismatch in future.
2. Having this package under ecosystem makes no help for users. They just do not realize we have such helpers at all. This even happens at default theme, I believe at least 1/5 people fails to find default theme docs in 3 minutes. So having a easy reachable `vuepress/bundler-helper` is enough as a motivation.
3. A lot of current vuepress users are developers using other languages as VitePress do not have such plugin systems. Bundlers are unfamilar concept for them (in their languages there just compiler and packer). So it helps that core wraps complex bundler options with helper functions.

meteorlxy:
Re1: That increases the complexity of core, which is not necessary.

Re2: That means we should improve the documentation, but not move it to core.

Re3: Users who do not familiar with bundlers should not update bundler options, should they?

IMO only advanced users / theme & plugin authors would try updating bundler options themselves, and they should be aware of the official ecosystem.

Mister-Hope:
> Re1: That increases the complexity of core, which is not necessary.

Again: so moving this to core ensures there is no support mismatch in future.

Avoiding possible future behavior mismatch is necessary. With a sub-export, we ensures our bundler helper is always up-to-date with current bundler.

> Re2: That means we should improve the documentation, but not move it to core.

Again: So having a easy reachable vuepress/bundler-helper is enough as a motivation.

The following looks like bullshit💩:

<details>
 <summarye>examples in official docs</summary>
 
```ts
export default {
  extendsBundlerOptions: (bundlerOptions, app) => {
    // extends options of @vuepress/bundler-vite
    if (app.options.bundler.name === '@vuepress/bundler-vite') {
      bundlerOptions.vuePluginOptions ??= {}
      bundlerOptions.vuePluginOptions.template ??= {}
      bundlerOptions.vuePluginOptions.template.compilerOptions ??= {}
      const isCustomElement =
        bundlerOptions.vuePluginOptions.template.compilerOptions.isCustomElement
      bundlerOptions.vuePluginOptions.template.compilerOptions.isCustomElement =
        (tag) => {
          if (isCustomElement?.(tag)) return true
          if (tag === 'my-custom-element') return true
        }
    }

    // extends options of @vuepress/bundler-webpack
    if (app.options.bundler.name === '@vuepress/bundler-webpack') {
      bundlerOptions.vue ??= {}
      bundlerOptions.vue.compilerOptions ??= {}
      const isCustomElement = bundlerOptions.vue.compilerOptions.isCustomElement
      bundlerOptions.vue.compilerOptions.isCustomElement = (tag) => {
        if (isCustomElement?.(tag)) return true
        if (tag === 'my-custom-element') return true
      }
    }

    // even more for the comming rspack
    // new shit here
  },
}
```

</details>

If VuePress core inisit leveling up the developing difficulty like this, it is expected that no new active contributors appears in our ecosystem.

We should not build high walls for developers and users. A lot of plugins WILL FAIL to run with vite/webpack with the default setting. 

I strongly inisit that our core should decrease the difficulty handling our bundler combinations with richer starter tools. The core SHOULD earse/lower the differences between bundlers just like how it's code is written. This should NOT be a ecosystem work.

I try to ask a lot of project which supports VitePress to support VuePress, and a lot of anwser is No, besides the reason that we are in RC stage,  a steedy understanding needs with our bundler system is the other major reason.

> Re3: Users who do not familiar with bundlers should not update bundler options, should they?

No, I've seen lots of reasonable needs comming from my theme users.

E.g.: 

- They imported a 3-rd party custom-component, that requires a customElement setting.
- They imported a heavy library, which requires manual chunking and preload/prefetch disabling (that's why I am still pushing effort to improve this option).


## Updated

Another major reason is that VuePress produces heavy/dirty bundelr config to make itself work. A careless override of any root config (e.g.: the whole `resolve` part) likely breaks how VuePress works.

At least the core are supposed to provide helpers like this:

```ts
addViteConfig(bundlerOptions, app, {
  // simplest option that is required
})

addWebpackConfig(bundlerOptions, app, {
  // simplest option that is required
})

addRspackConfig(bundlerOptions, app, {
  // simplest option that is required
})
```

We are providing a documentation tool, not a devloping tool, so it's faire that we requires the core to care adjusting bundling behavior.

PR #1709, opened 2026-05-19
feat(bundler-vite): add configureVite option

Mister-Hope:


Mister-Hope:
This should be merged before the bundelr helper.

packages/bundler-vite/src/types.ts
```diff
@@ -8,4 +8,9 @@ import type { InlineConfig } from 'vite'
 export interface ViteBundlerOptions extends BundlerOptions {
   viteOptions?: InlineConfig
   vuePluginOptions?: VuePluginOptions
+  configureVite?: (
```
meteorlxy, line 25: Please help add comments for each options.

Mister-Hope: Done

Mister-Hope:
https://github.com/vuepress/core/pull/1714 now replaces this PR with a full refine.

However, if multiple pr is preferred, merge this first.

packages/bundler-vite/src/types.ts
```diff
@@ -6,6 +6,25 @@ import type { InlineConfig } from 'vite'
...
+   * Options for @vitejs/plugin-vue
+   */
   vuePluginOptions?: VuePluginOptions
+  /**
+   * Modify Vite config
+   *
+   * @param config - Vite config
+   * @param isServer - Whether it is server bundle
+   * @param isBuild - Whether in build mode
+   * @returns Vite config to be merged with the default config
```
meteorlxy, line 23: Need to clarify that an empty return is also valid.

Mister-Hope: See https://github.com/vuepress/core/pull/1714 part 3. We do not allow "Vite config to be merged with the default config" anymore in that.

This msg will be updated to a new one after that is merged.

In short, 1714 covers all things in this PR, just because you do not like heavy refactors in other contributors' PRs rather than you.

I don't hink any futher update is needed, this msg will be removed anyway after the whole improvement.

Mister-Hope: In short, you should review #1714 first, this PR just behaves like a refactor splitting.

If you want, you can merge with rebase for that, and this will no longer needed.

Please leave your futher opinion there.

PR #1711, opened 2026-05-19
feat(core)!: add as-needed prefetch strategy

### LINKED ISSUE #952: [Feature request] Improve shouldPrefetch and shouldPrefetch option

Currenly the `(file: string, type: string) => boolean` is TOO unmeaningful.

The file args in vite is the output chunk name, which is probably like `index.html.xxxx.js`.

image

It lost infoamtion about how it was generated, so users at this time have No idea about what's page is this chunk generated from. So it's impossible for a developer or a user to take actual effect when setting a function.

Just an example, I am getting hundreds of `index.html.xxxx.js` in https://github.com/vuepress-theme-hope/v2/tree/gh-pages/assets.

image

### Suggested solution

The parameters should be rebuilt. At least it should include:

- chunk filename
- chunk  filetype (can be ommited if filename contains extension). Also I am confused why current api seperate a full chunk path into file name and file type.
- chunk infomation about how this chunk is generated.

   Could be things like:

```js
{from: 'markdown', source: 'path/to/markdown.md'}
```

```js
{from:'import', source: 'path/to/files/where/the/chunk/is/imported.ts'}
```

```js
{from:'script', source: 'path/to/file/entry.ts'}
```

Edited: In some cases we might get an array here, so we can use array formats.

### Alternative

_No response_

### Additional context

_No response_

### LINKED ISSUE #953: [Feature request] Improve shouldPrefetch default value

Currenly the `shouldPrefetch: true` default behavior have serious preformance impact on large sites with FCP time.

For a large site with around 1500 pages, shouldPrefetch link can take up to 30-50KB in each html entry.

### Suggested solution

As @meteorlxy suggests earlier, we can change the default behavior to only prefetch assets which can be visited through the page.

Or more detailed:

1. find all links in the page, (also home and 404)
2. find all links in navbar and sidebar
3. Combine them together with a link array
4. Get all the preload links of them
5. filter the duplicated ones
6. Set them as default value of `shouldPretch`

But this requires improvments mentioned in https://github.com/vuepress/vuepress-next/issues/952

### Alternative

_No response_

### Additional context

_No response_

Mister-Hope:


Mister-Hope:
@meteorlxy This is completed, but we should expected the following things to be preloaded as well:

- pages appears in navbars
- pages appears in the same sidebar group
- pages that is prev/next

But these are theme concept, and I am not comming up with any idea how to solve this gracefully.

Any ideas for probably a new hook for themes?

packages/bundler-vite/src/build/build.ts
```diff
@@ -65,6 +67,15 @@ export const build = async (
     )
     const ssrTemplate = await getSsrTemplate(app)
 
+    // build page path -> chunk files map for 'as-needed' strategy
```
meteorlxy:
```suggestion
    // build page path -> chunk files map for preload & prefetch
```

packages/bundler-webpack/src/build/build.ts
```diff
@@ -78,6 +79,16 @@ export const build = async (
     const { initialFilesMeta, asyncFilesMeta, moduleFilesMetaMap } =
       resolveClientManifestMeta(clientManifest)
 
+    // build page path -> chunk files map for 'as-needed' strategy
```
meteorlxy:
```suggestion
    // build page path -> chunk files map for preload & prefetch
```

packages/bundler-vite/src/build/renderPagePrefetchLinks.ts
```diff
@@ -21,12 +27,37 @@ export const renderPagePrefetchLinks = ({
...
   }
 
-  // dynamic imports excluding current page chunks
-  const prefetchFiles = outputEntryChunk.dynamicImports.filter(
-    (item) => !pageChunkFiles.some((file) => file === item),
-  )
+  let candidateFiles: string[]
 
-  return prefetchFiles
+  if (shouldPrefetch === 'as-needed') {
```
meteorlxy, line 32: Maybe extract this snippet to bundlerutils?

Mister-Hope: How? The data structures are different accross bundlers, and rely on bundler types.

meteorlxy: L33-L45 is identical and duplicate across 3(4) files

meteorlxy: extracted to bundlerutils

meteorlxy (approved):

PR #1714, opened 2026-05-21
feat(core)!: refine bundler options

Mister-Hope:
## Summary

Refine the `Bundler` interface and bundler configuration hooks to give plugin authors the tools they need to configure bundler-specific behavior correctly and completely.

---

### 1. Complex comparison to identify the active bundler

Plugins in `extendsBundlerOptions` need to know which bundler is active to add the right options (e.g. `viteOptions` vs `chainWebpack`). The only discriminator was `bundler.name`, which is the full scoped package name.

```diff
- app.options.bundler.name === '@vuepress/bundler-webpack'
+ app.options.bundler.type === 'webpack'
```

### 2. Missing merge helper in `extendsBundlerOptions`

When a plugin accepts user bundler options and wants to merge them into the resolved options, it needs the correct merge function (`vite.mergeConfig` vs `webpack-merge` differ in semantics). The hook only passes `(options, app)` — no merge helper.

We considered adding `mergeConfig` as a hook parameter, but all `extends*` hooks share the uniform signature `(extendable: T, app: App)`:

```typescript
// hooks.ts — every extends hook follows this pattern
extendsMarkdownOptions: ExtendsHook<MarkdownOptions>
extendsMarkdown: ExtendsHook<Markdown>
extendsPageOptions: ExtendsHook<PageOptions>
extendsPage: ExtendsHook<Page>
extendsBundlerOptions: ExtendsHook<BundlerOptions>
```

Adding a third parameter would break this consistency. So `mergeConfig` is exposed on the `Bundler` object instead:

```diff
  export interface Bundler {
    name: string
+   type: string
    dev: (app: App) => Promise<() => Promise<void>>
    build: (app: App) => Promise<void>
+   mergeConfig: <Config extends Record<string, any>>(
+     currentConfig: Config,
+     newConfig: DeepPartial<Config>,
+   ) => Config
  }
```

Plugins access it naturally: `app.options.bundler.mergeConfig(...)`.

### 3. `configureWebpack` / `configureVite` cannot remove array items

The old merge-on-return model was purely additive — array items could only be appended, never deleted.

```diff
  // resolveWebpackConfig.ts
  const configureWebpackResult = options.configureWebpack?.(...)

- // if `configureWebpack` returns a configuration object,
- // use webpack-merge to merge it
  if (configureWebpackResult) {
-   return merge(webpackConfig, configureWebpackResult)
+   webpackConfig = configureWebpackResult
  }
```

Returning a value now **fully replaces** the config. ~To merge explicitly, use the `mergeConfig` callback parameter passed to the hook.~

Note: the `mergeConfig` from vite always create a fresh new object at every level while merging.

---

## Breaking changes

- **`Bundler` interface** now requires `type` and `mergeConfig`.
- **`configureWebpack`** return value replaces the config. Use the passed `mergeConfig` parameter to merge instead.


packages/bundler-vite/src/types.ts
```diff
@@ -6,6 +6,36 @@ import type { InlineConfig } from 'vite'
...
+  /**
+   * Vite options
+   */
   viteOptions?: InlineConfig
+  /**
+   * Options for @vitejs/plugin-vue
+   */
+  /**
+   * Options for @vitejs/plugin-vue
+   */
```
Mister-Hope:
```suggestion
  /**
   * Options for @vitejs/plugin-vue
   */
```

packages/bundler-vite/src/types.ts
```diff
@@ -6,6 +6,36 @@ import type { InlineConfig } from 'vite'
  * Options for bundler-vite
  */
 export interface ViteBundlerOptions extends BundlerOptions {
+  /**
+   * Vite options
+   */
+  /**
+   * Vite options
+   */
```
Mister-Hope:
```suggestion
  /**
   * Vite options
   */
```

packages/bundler-vite/src/types.ts
```diff
@@ -6,6 +6,30 @@ import type { InlineConfig } from 'vite'
...
   vuePluginOptions?: VuePluginOptions
+  /**
+   * Modify Vite config
+   *
+   * @param config - Vite config
+   * @param isServer - Whether it is server bundle
+   * @param isBuild - Whether in build mode
+   * @param mergeConfig - Vite's mergeConfig function, in case you want to use it to merge config
+   * @returns New Vite config
```
Mister-Hope:
```suggestion
   * @returns optional whole new Vite config
```

packages/bundler-webpack/src/types.ts
Mister-Hope, line 28:
```suggestion
/**
  * use webpack-merge to set webpack config
  * 
  *
  * @param config - Vite config
  * @param isServer - Whether it is server bundle
  * @param isBuild - Whether in build mode
  * @param mergeConfig -  merge function from webpack-merge, in case you want to use it to merge config
  * @returns optional whole new webpack config
  */
```

packages/bundler-webpack/src/types.ts
```diff
@@ -20,12 +20,25 @@ export type {
 export interface WebpackBundlerOptions extends BundlerOptions {
   /**
    * use webpack-merge to set webpack config
-   */
+/**
+  * use webpack-merge to set webpack config
+  * 
+  *
+  * @param config - Vite config
```
Mister-Hope:
```suggestion
  * @param config - Webpack config
```

meteorlxy:
- feat(bundler-vite): add `configureVite` option
- feat(bundler-vite): export `viteMergeConfig` function
- feat(bundler-webpack): export `webpackMergeConfig` function
- feat(core): add `type` field to bundlers to indicate the underlying bundler
- feat(core): add `mergeConfig` field to bundlers to provide config merging helper

BREAKING CHANGE: core - `Bundler `interface now requires `type` and `mergeConfig`.
BREAKING CHANGE: bundler-webpack - `configureWebpack` return value now replaces the config object directly. Use the `webpackMergeConfig` helper to merge configs manually instead.

meteorlxy (approved):

